Issue #5, opened 2017-11-17
Create authentication workflow


dereklieu:
This will just authenticate against the Django api for now. New user creation will punt until we have a working email notification system.

PR #8, opened 2017-11-17
Feature/auth connection

dereklieu:
Contributes to #5 

- Creates `get` function to abstract away the hassle of dispatching `-inflight`, `-failed`, and `-success` actions.
- Creates a listener as middleware, to save token to localstorage when there is a successful token query.
- Creates a `withToken` function to add the `Authorization` header and include token from localstorage to fetch options.
- Adds actions to trigger fetches for tokens, and field reports by `pk`.

One TODO is to check the token `expires` property on boot and if it's expired, delete it. Leaving that up to you @danielfdsilva as it plays into the whole login workflow.

Currently there's no `logout` endpoint on the API, I'm not sure we need to support one either but open to creating it if you think it's necessary.

Currently the deployed API doesn't support this feature yet, but I will ping here when it does (and the database has been populated).


danielfdsilva:
Fixes #5 
- Add a user page (dumps data for now)
- Add a user menu to view account and logout
- Restricts access to page based on authentication
- Refactor code

Todo:
- [ ] Handle login error (depends on api's response being json)

@dereklieu you can review work so far

PR #11, opened 2017-11-23
Post field report

dereklieu:
@danielfdsilva This adds a dummy page that requests the field report schema, requests a hard-coded field report, and includes a fixture that you can instantly `POST` to the api. It also includes all the action creators and state objects you need to do that.

On creating records, apparently the framework we're using will accept input more or less the way it serves it, so for future reference doing a read on a current record should be a good indicator for how to structure the payload.

When you get to rendering forms for the Countries selector and the Disaster Type selectors, the data for those can be found here: https://github.com/IFRCGo/go-api/tree/develop/api/fixtures

Those are unlikely to change so I would just pull them into this codebase to use to generate dropdowns or whatever other form element we need.

For this particular model, many of the fields are optional. For the few that aren't optional, either consult the schema, or check the model here: https://github.com/IFRCGo/go-api/blob/develop/api/models.py#L73. `null=True` or `blank=True` means it's optional.

### DIFF
--- a/app/assets/scripts/actions/index.js
+++ b/app/assets/scripts/actions/index.js
@@ -1,9 +1,13 @@
 'use strict';
-import { fetchJSON, withToken } from '../utils/network';
+import {
+  fetchJSON,
+  withToken,
+  createResource
+} from '../utils/network';
 
 export const TOKEN = 'TOKEN';
 export function getAuthToken (username, password) {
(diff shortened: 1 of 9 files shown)


dereklieu:
@danielfdsilva feel free to close this and delete if/when this branch loses relevance to your feature branch. I just made it as a demo.

PR #12, opened 2017-11-23
Field report form

danielfdsilva:


app/assets/scripts/utils/field-report-constants.js
```diff
@@ -0,0 +1,150 @@
...
+    description: 'Early warning message for an upcoming disaster.'
+  },
+  {
+    label: 'Event',
+    value: 'event',
+    description: 'First report for this disaster.'
+  }
+];
+
+export const countries = [
```
dereklieu: I know these are just here for the sake of development, but would it be helpful to have endpoints for all of these (ie countries, disaster types, etc)? Some of them, like `Events`, will need API endpoints of course.

app/assets/scripts/views/field-report-form.js
```diff
@@ -0,0 +1,1235 @@
...
+      this.setState({ step: this.state.step - 1 });
+    }
+  }
+
+  onStepperClick (step, e) {
+    e.preventDefault();
+    // Can't move forward. Only backwards.
+    // if (step > this.state.step) {
+    //   return;
+    // }
```
dereklieu: This is probably Ok to leave out completely yeah? People will be able to hop around if they wish.

app/assets/scripts/views/field-report-form.js
```diff
@@ -0,0 +1,1235 @@
...
+      'Basic Information',
+      'Numeric Details',
+      'Actions Taken',
+      'Planned Response',
+      'Contacts'
+    ];
+    return (
+      <ol className='stepper'>
+        {items.map((o, idx) => {
+          idx += 1;
```
dereklieu: I was confused about this construction for a bit, not realizing it doesn't manipulate the actual loop. Maybe a separate variable could make this more clear?

dereklieu:
@danielfdsilva This is working great. One suggestion to maybe consider is to either scroll the page up or down on a validation error, or have some other way of showing the validation errors, since right now it can be hard to know what the errors are (they render below the bottom of the page for me).

A likewise suggestion would be to scroll to the top of the form after completing a section.

app/assets/scripts/views/field-report-form.js
```diff
@@ -0,0 +1,1235 @@
...
+            <FormError
+              errors={this.state.errors}
+              property='disasterType'
+            />
+          </FormSelect>
+
+          <FormSelect
+            label='Event'
+            name='event'
+            id='event'
```
dereklieu: Eventually (very soon hopefully) this can turn into an autocomplete form that will query the API. Obviously we need to keep it like this for now, but just to keep that in mind.

dereklieu:
@danielfdsilva feel free to merge!

app/assets/scripts/utils/field-report-constants.js
danielfdsilva: Ideally that would be good yes, but given the development effort I think that implementation will depend on how likely is this to change. Countries for example, shouldn't change, so it might be ok to keep as a constant.

app/assets/scripts/views/field-report-form.js
danielfdsilva: This is a behavior we have to decide on. Even if hopping around we should ensure that the page they're in is valid. They shouldn't be able to move around if there are errors in the page. no?

danielfdsilva:
@dereklieu Made some improvements:
- Scroll to top on step change
- Display system alert when there are errors in the form.

Still todo: (will open an issue for this)
- Improve error messages text
- Style error summary at the bottom
- Scroll to error (enhancement)

dereklieu:
@danielfdsilva the new changes look good but running this, I get an error because `<SysAlerts />` is being remounted and throwing an error. This happens whenever I navigate to a new page.

I pushed this proposed fix in #15, if that fixes the issue adequately than feel free to merge this in.

danielfdsilva:
@dereklieu had the exact problem with `GlobalLoading` but I guess I missed this one! Thanks! Merging everything

PR #29, opened 2017-12-05
[WIP] Homepage

danielfdsilva:


app/assets/scripts/actions/index.js
```diff
@@ -26,6 +28,35 @@ export function createFieldReport (payload) {
...
+  const f = buildAPIQS(filters);
+
+  return fetchJSON(`/api/v1/surge_alert/?${f}`, GET_SURGE_ALERTS, withToken());
+}
+
+export const GET_APPEALS_LIST = 'GET_APPEALS_LIST';
+export function getAppealsList () {
+  const f = buildAPIQS({
+    end_date__gt: DateTime.local().toISODate(),
+    limit: 100
```
dereklieu: What's your plan for this, and the recursive function below, with real data? I feel like setting the limit to `0` should be ok now that we're using reasonable development data?

app/assets/scripts/components/connected/alerts-table.js
```diff
@@ -0,0 +1,161 @@
...
+    }
+  }
+
+  requestResults () {
+    this.props._getSurgeAlerts(this.state.page);
+  }
+
+  handlePageChange (page) {
+    this.setState({ page: page.selected + 1 }, () => {
+      showGlobalLoading();
```
dereklieu: Could just put `showGlobalLoading` in the `requestResults` function.

app/assets/scripts/utils/country-centroids.js
```diff
@@ -0,0 +1,247 @@
...
+  WS: [ -172.11331132714278, -13.786946593957028 ],
+  YE: [ 47.1086302234567, 14.56738963876404 ],
+  ZA: [ 24.703232827543786, -29.26925840866997 ],
+  ZM: [ 28.489715084919187, -13.149159999209038 ],
+  ZW: [ 29.389317242443635, -18.5589791387905 ]
+};
+
+export function getCentroid (iso) {
+  // Centroid or null island.
+  return centroids[iso.toUpperCase()] || [0, 0];
```
dereklieu: We might want to have something here that will prevent mapping things to `[0, 0]`. Maybe returning false would make it easier to catch those?

dereklieu:
Overall this is coming along really well. The map data isn't showing up for me though, is it supposed to?

app/assets/scripts/actions/index.js
danielfdsilva: It's just a matter of removing the recursion block.

app/assets/scripts/utils/country-centroids.js
danielfdsilva: What should be the desired behavior? If we have no centroid don't show the data?
I used null island to present the app from breaking.

danielfdsilva:
@dereklieu map should be showing data.
image

PR #34, opened 2017-12-06
User notification subscriptions

dereklieu:
- Adds forms for controlling user subscriptions
- Creates an action to post these preferences to the API

TODO:
- [x] Add reducer state to record the success/failure of the request to update user subscriptions

@danielfdsilva I'm still wrapping up work on the API side to support this, but the front-end side could be ready for a quick read if you've got a moment.

### DIFF
--- a/app/assets/scripts/views/account.js
+++ b/app/assets/scripts/views/account.js
@@ -132,11 +132,11 @@ class Account extends React.Component {
   }
 
   syncNotificationState (data) {
-    const subscriptions = _get(data, 'subscription', [])
+    const subscriptions = _get(data, 'subscription', []);
     if (!subscriptions.length) {
       return;
     }
-    let next = Object.assign({}, this.state.data)
+    let next = Object.assign({}, this.state.data);
     subscriptions.forEach(sub => {
       const rtype = rtypes[sub.rtype];
       if (rtype === 'country' && sub.country) {
--- a/app/assets/scripts/views/account.js
+++ b/app/assets/scripts/views/account.js
@@ -7,7 +7,6 @@ import _set from 'lodash.set';
 import _cloneDeep from 'lodash.clonedeep';
 import _get from 'lodash.get';
 import c from 'classnames';
-import createFragment from 'react-addons-create-fragment';
 
 import { environment } from '../config';
 import {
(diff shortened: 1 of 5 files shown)


dereklieu:
@danielfdsilva Ok, this is wrapped up. If you want to give it a review tomorrow, the API should also be synced by then.

app/assets/scripts/views/account.js
```diff
@@ -2,40 +2,290 @@
...
+      'last_name',
+      'profile.city',
+      'profile.org',
+      'profile.org_type',
+      'profile.department',
+      'profile.position',
+      'profile.phone_number'
     ];
-    const profileData = profile.data.objects[0];
     return attributes.map(a => [
```
danielfdsilva: Can you use `React.Fragment` instead of an array? :)
The shorter syntax is not yet supported, but with the name works fine!

app/assets/scripts/views/account.js
```diff
@@ -2,40 +2,290 @@
...
+            values={this.state.data.appeal}
+            onChange={this.onFieldChange.bind(this, 'appeal')} />
+          <FormCheckboxGroup
+            label='Other Notifications'
+            name='other'
+            classWrapper='action-checkboxes'
+            options={otherNotificationTypes}
+            values={this.state.data.other}
+            onChange={this.onFieldChange.bind(this, 'other')} />
+          <button type='submit' className={c('button', {
```
danielfdsilva: Should the button be disabled if not dirty?

dereklieu:
@danielfdsilva ok, I switched to using the `disabled` class and also added the create fragment addon library. Let me know if [this](https://github.com/IFRCGo/go-frontend/pull/34/commits/fd18383db842bfceaa7ea38e796197a48e01d162#diff-e268a725e3f566f0c27f8234072d2b35R218) is not the format you had in mind.

danielfdsilva:
@dereklieu Apologies, should have explained better. :/
Fragments are now part of React! Yay!

you can do:
```
<React.Fragment key={a}>
  <dt>{apiPropertyDisplay(a)}</dt>
  <dd>{apiPropertyValue(a, profile.data)}</dd>
</React.Fragment>
```

dereklieu:
@danielfdsilva no worries, the last commit should do it. It required a newer React version, and that version got bumped in the lockfile but not package.json, which I think is fine.

danielfdsilva:
@dereklieu ups. I upgraded on the homepage branch. Thought that those changes were in already.
Anyway, looks good. merging

PR #37, opened 2017-12-12
[WIP] Feature/emergencies

ascalamogna:


dereklieu:
Map component changes:

1. Moves map init to `utils/get-new-map` (aka setting map drag settings, etc.)
1. Adds an abstract map component to `components/map/map`.
1. Abstract map component takes `geojson, filters, layers` props as a way to drive rendering.
1. Parent map components are still responsible for setting event listeners and opening popups, and get a `setupListeners` hook to do this.

In the abstract map component, difference checking is done using object-level checks aka `thisFilters === nextFilters` so we just need to make sure to create new `filter`, `layer`, and `geoJSON` objects rather than extending current ones, but this is good practice anyway and already happening.

app/assets/scripts/components/connected/emergencies-table.js
```diff
@@ -0,0 +1,162 @@
...
+      return (
+        <Fold title='Latest Alerts'>
+          <p>Oh no! An error ocurred getting the stats.</p>
+        </Fold>
+      );
+    }
+
+    if (!data.objects.length) {
+      return (
+        <Fold title='Latest Alerts'>
```
danielfdsilva: Here and above you have "Latest Alerts" but below you have "Latest Emergencies". Is this correct?

dereklieu: 👍 they should all be "Latest Emergencies"

app/assets/scripts/reducers/emergencies.js
```diff
@@ -0,0 +1,190 @@
...
+      countries[c.iso].records.push(record);
+    });
+  });
+
+  const geoJSON = {
+    type: 'FeatureCollection',
+    features: Object.keys(countries).map(iso => {
+      const { country, records } = countries[iso];
+
+      const properties = {
```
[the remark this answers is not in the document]
dereklieu: 👍

dereklieu:
@danielfdsilva I've incorporated your feedback and synced with `develop`. This should be good to go, pending one last review.

PR #38, opened 2017-12-12
Feature/sync field report

dereklieu:
> When pushing a foreign key relation, it should be pushed as a uri, aka `api/v1/country/{pk}`.

Note that this wasn't necessary, the simpler form is just `{pk: pk}`


danielfdsilva:
@dereklieu Looks good!
Maybe just check the comments on the beginning of the file, and delete what's done.

dereklieu:
@danielfdsilva this is now working. I've deployed the necessary changes to the API as well. Let me know if you have any suggestions.

app/assets/scripts/views/field-report-form.js
```diff
@@ -297,18 +260,21 @@ class FieldReportForm extends React.Component {
...
-        console.log('nextProps.fieldReportForm', nextProps.fieldReportForm);
-        hideGlobalLoading();
+    if (this.props.fieldReportForm.fetching && !nextProps.fieldReportForm.fetching) {
+      hideGlobalLoading();
+      if (nextProps.fieldReportForm.error) {
+        const message = nextProps.fieldReportForm.error.error_message || 'Could not submit field report';
+        showAlert('danger', <p><strong>Error:</strong> {message}</p>, true, 4500);
+      } else {
+        const { id } = nextProps.fieldReportForm.data;
+        showAlert('success', <p><Link to={`/reports/${id}`}>Field report created</Link></p>, true, 9000);
```
danielfdsilva: Maybe this alert shouldn't autodismiss?

app/assets/scripts/views/field-report-form.js
```diff
@@ -469,9 +446,15 @@ class FieldReportForm extends React.Component {
...
       if (step === 5) {
-        console.log('Submit data!!!', this.getSubmitPayload());
-        // this.props._createFieldReport(this.getSubmitPayload());
-        // showGlobalLoading();
+        const payload = this.getSubmitPayload();
+        const userId = _get(this.props.user, 'data.id');
+        if (userId) {
+          payload.user = {pk: userId};
+        } else {
+          console.log('Could not read user ID from state');
```
danielfdsilva: Does it still work without user Id or should this be handled as an error?

dereklieu: It would work without a user ID, but that should never happen since this is a logged-in route.

app/assets/scripts/views/field-report-form.js
dereklieu: @danielfdsilva actually I think this should redirect to the field report page. If you click save again it will create a duplicate record, which is not great.

PR #46, opened 2017-12-15
Fix sass build

dereklieu:
Sass was complaining about `$cycle: none` on a production build. Setting `0` instead of `none` worked. Dunno what this does so @ascalamogna or @danielfdsilva could you weigh in?


danielfdsilva:
@dereklieu this was fixed by me in another branch.
It's already in develop (https://github.com/IFRCGo/go-frontend/blob/develop/app/assets/styles/global/_pages.scss). This diff is actually weird.

Please double check and close

dereklieu:
👍

PR #49, opened 2017-12-15
Include mapping of country ids to region ids

dereklieu:
@danielfdsilva you'll need this utility in order to query for all of the deployed resources within a *region*.

Currently those resources only have a `country` property. Adding an additional `region` property would make data entry very inconvenient over the long run.

Here are the current region IDs: https://github.com/IFRCGo/go-api/blob/develop/api/models.py#L36

(MENA = Middle East & North Africa)

So if you wanted to search for all ERUs deployed in MENA, your query would be `api/v1/eru/countries__in=2,10,16,26,61,87,...`

Each record in the result will look like this:

screen shot 2017-12-15 at 5 09 01 pm

`countries` is where the resource is deployed (which would include one of the countries in your query). The `eru_owner` is the national society country that owns the resource. `type` is the ERU type, which is defined [here](https://github.com/IFRCGo/go-api/blob/develop/api/models.py#L454). `units` is just the number of ERU units deployed.


danielfdsilva:
This is ready for a review.
There are some components that are the same as the homepage, and maybe will be used on the countries page as well, so we should look into making components out of them.
I feel like there's a bit too much replication going on here.

@derek thoughts??

dereklieu:
@danielfdsilva yeah I generally agree. What components in particular are you referring to though? 

Given the timeline I'm more inclined to finish the country page and repeat some components, then come back and refactor later. If there's low-hanging fruit though than I'd be open to it.

app/assets/scripts/reducers/region.js
```diff
@@ -0,0 +1,245 @@
...
+  });
+};
+
+function regionData (state = initialState, action) {
+  switch (action.type) {
+    case 'GET_REGION_INFLIGHT':
+      state = Object.assign({}, state, {
+        [action.id]: stateInflight(state, action)
+      });
+      state = stateInflight(state, action, action.id);
```
dereklieu: What does this second state assignment do?

app/assets/scripts/views/region.js
```diff
@@ -0,0 +1,597 @@
...
+          { id: 'event', label: 'Event' },
+          { id: 'dtype', label: 'Disaster Type' },
+          { id: 'requestAmount', label: 'Appeal Amount (CHF)' },
+          { id: 'fundedAmount', label: 'Funding (CHF)' },
+          { id: 'active', label: 'Active' }
+        ];
+
+        const rows = data.objects.map(o => ({
+          id: o.id,
+          date: DateTime.fromISO(o.created_at).toISODate(),
```
dereklieu: I think for appeals and drefs, either `start_date` or `end_date` would be a better proxy for this than created_at, which will all be the same right now. If I were to pick one it'd probably be `end_date`.

app/assets/scripts/actions/index.js
```diff
@@ -98,3 +100,80 @@ export const GET_EVENT = 'GET_EVENT';
...
+      }
+    }, 1);
+  };
+}
+
+export const GET_REGION_APPEALS = 'GET_REGION_APPEALS';
+export function getRegionAppeals (regionId, page = 1, filters = {}) {
+  filters.limit = filters.limit || 10;
+  filters.offset = filters.limit * (page - 1);
+  filters.atype = 0;
```
dereklieu: This should be `1`.

dereklieu:
Looks good, let's merge after you check out the inline comments.

app/assets/scripts/views/region.js
danielfdsilva: Changed for appeals and drefs. What about field reports?

app/assets/scripts/reducers/region.js
danielfdsilva: Nothing. that's a mistake

danielfdsilva:
@dereklieu agreed. let's finish this and then refactor if we have time

PR #57, opened 2017-12-20
HeOps page

ascalamogna:


dereklieu:
So far done:

- Create action and reducers for the two charts and the table on this page
- Hooked up the data for the table (but it still needs the pagination element).

@danielfdsilva if you'd like to continue this tomorrow morning feel free to do so - otherwise I'll finish it when I start.

dereklieu:
This is done. It was built entirely with existing components. Right now it doesn't look great because the test data isn't very rich, but once that gets filled in, this should come together as well.

screen shot 2017-12-24 at 8 33 13 pm

screen shot 2017-12-24 at 8 33 17 pm

PR #59, opened 2017-12-21
[WIP] Country page

danielfdsilva:


app/assets/scripts/views/admin-area.js
```diff
@@ -491,7 +491,7 @@ class AdminArea extends React.Component {
 
   render () {
     return (
-      <App className='page--region'>
+      <App className='page--country'>
```
danielfdsilva: @ascalamogna This component is used both for the country and the region. Could we use `page--admin-area` as a class so it's clearer and works for both?

ascalamogna: @danielfdsilva there's specific styles needed for the country page, I modified the layout a bit so it needs a separate class.

danielfdsilva:
Ok. I'll make the class dynamic then, otherwise the region page has the
same class

danielfdsilva:
@ascalamogna fixed. Please test and merge at will. :)

Issue #58, opened 2017-12-20
Add Sort/Filter Functionalities to tables


ascalamogna:
**Emergencies**
date - dropdown 
Name - sort alphabetical
Disaster Type - dropdown
total affected - sort 
total beneficiaries  - sort

**Field Reports**
date -dropdown
disaster type- dropdown
region - dropdown 

**Appeals/Drefs**
date -dropdown
name - alphabetical 
disaster type -dropdown
requested funding - sort
funding - sort 

**HeOps**
region -dropdown
name-dropdown
role- dropdown

For date let's do the following ranges unless some has better suggestions:
- last week
- last month
- last year

cc @dereklieu @danielfdsilva

danielfdsilva:
Pretty much done

PR #68, opened 2017-12-22
Sort and filter

danielfdsilva:
This PR adds a table comoponent which allow to easily set filters and sorts on the headers.
Check the `admin-area.js` file to see how it's implemented.

TODO on the `admin-area.js`
- [x] Compute filters
- [x] Verify api connection

TODO overall;
- [x] Replace tables with Component. There are some pages where this table structure is not being used, like the homepage. Changing it should be pretty fast.

Any question just reach out

cc @derek

### DIFF
--- a/app/assets/scripts/components/connected/alerts-table.js
+++ b/app/assets/scripts/components/connected/alerts-table.js
@@ -3,13 +3,15 @@ import React from 'react';
 import { connect } from 'react-redux';
 import { PropTypes as T } from 'prop-types';
 import { Link } from 'react-router-dom';
-import ReactPaginate from 'react-paginate';
 import { DateTime } from 'luxon';
 
 import { environment } from '../../config';
 import { getSurgeAlerts } from '../../actions';
-import BlockLoading from '../block-loading';
+import { dateOptions, datesAgo } from '../../utils/utils/';
 
+import { SFPComponent } from '../../utils/extendables';
+import DisplayTable, { FilterHeader } from '../display-table';
+import BlockLoading from '../block-loading';
 import Fold from '../fold';
 
 const alertTypes = {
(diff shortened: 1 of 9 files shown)


danielfdsilva:
@dereklieu I'd say that overall this is ready. Just 2 things missing:

- [ ] Appeals, DREFs and Field Reports are not filterable by dtype
- [ ] Alerts table not filterable by date

There may be missing headers to sort / filter. We should do a general check.

danielfdsilva:
@dereklieu @ascalamogna Reviewed the sort/filter according to #58.

**Emergencies**
date - dropdown
> For the date we're displaying `disaster_start_date` or `start_date` or `created_at` depending on what is available. @dereklieu How do you propose we implement a filter here?

Disaster Type - dropdown
> @dereklieu is checking feasibility of filtering by a relation

total affected - sort
> This is not present on the table

total beneficiaries - sort
> This is a sum of the beneficiaries of each appeal. @dereklieu can this be sorted?

**Field Reports**
disaster type- dropdown
> @dereklieu is checking feasibility of filtering by a relation

region - dropdown
> This is not present on the table

**Appeals/Drefs**
disaster type -dropdown
> @dereklieu is checking feasibility of filtering by a relation

**HeOps**
> I used the implemented structure as guide for what to sort/filter on.

dereklieu:
@danielfdsilva @ascalamogna it turns out that you can directly order by related fields such as `country` or `dtype` and it works as expected. Aggregated fields, such as `total_beneficiaries` for the emergencies page, which is calculated from all the associated appeals, cannot be ordered on, and neither can many-to-many fields such as `regions` and `countries.`

dereklieu:
@danielfdsilva also for the emergencies date, I propose we use `disaster_start_date`.

PR #80, opened 2018-01-03
add buttons for saving field report

ascalamogna:


app/assets/scripts/views/field-report-form/index.js
```diff
@@ -748,8 +748,9 @@ class FieldReportForm extends React.Component {
                 {this.renderErrorSummary()}
 
                 <div className='form__actions'>
-                  <button type='button' className={c('button button--secondary-bounded', {disabled: this.state.step <= 1})} title='Go back to previous step' onClick={this.onStepBackClick}>Back</button>
+                  <button type='button' className={c('button button--base-plain', {disabled: this.state.step <= 1})} title='Go back to previous step' onClick={this.onStepBackClick}>Back</button>
                   <button type='submit' className='button button--secondary-filled' title='Save and continue'>Save and continue</button>
+                  <button type='submit' className='button button--secondary-bounded' title='Save and continue'>Save as Draft</button>
```
dereklieu: @ascalamogna I think we should avoid having two `type=submit` elements, as that could cause weirdness or unexpected behavior when you hit enter in the form. Also I think the title of the draft button needs to be changed.

guipizzini:
@dereklieu can this be pulled through?

dereklieu:
@guipizzini we haven't had a chance to wrap up API support for this yet. Will try to fit that in if possible but it may slide to phase 2.

PR #87, opened 2018-01-05
[WIP] IE11 support

danielfdsilva:
There are 2 major problem supporting IE11.
First one react fragments are not working (https://github.com/facebook/react/issues/11759). This was fixed and they should have a release in a week or so.

Second problem is styling. We're using flexbox for many things which is not incredibly compatible with IE11. Maybe something can be done? cc @ascalamogna 

Besides this I didn't find any other problems.

### DIFF
--- a/app/assets/scripts/main.js
+++ b/app/assets/scripts/main.js
@@ -107,3 +107,21 @@ render(
   <Root store={store} />,
   document.querySelector('#app-container')
 );
+
+// Polyfill for HTML Node remove();
+// https://developer.mozilla.org/en-US/docs/Web/API/ChildNode/remove
+(function (arr) {
+  arr.forEach(function (item) {
+    if (item.hasOwnProperty('remove')) {
+      return;
+    }
+    Object.defineProperty(item, 'remove', {
+      configurable: true,
+      enumerable: true,
+      writable: true,
+      value: function remove () {
+        if (this.parentNode !== null) this.parentNode.removeChild(this);
+      }
+    });
+  });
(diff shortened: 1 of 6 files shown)


dereklieu:
Facebook still hasn't cut a release. @danielfdsilva I think it would be worth spending a half hour to see if it's feasible to build React from current master, and bundling it with our repo, at least until the new version is released. Wouldn't spend more than that if it gets complicated. What do you think?

danielfdsilva:
@dereklieu Installed this version of react (https://www.npmjs.com/package/@xavescor/react) which is listed on the issue. Also rebased from develop to keep it up to date.
It is now working on IE. The styles are still breaking. cc @ascalamogna

dereklieu:
@danielfdsilva :+1: could we include an `ie` class to one of the top-level parent containers, so that we can apply the appropriate styles for this browser?

@ascalamogna We're supporting IE 11 to the extent that it doesn't look broken. If we need to hide features, show fewer table columns, or revert to mobile styles, that's totally within reason.

danielfdsilva:
@dereklieu Added IE detection. However IE11 supports most styles easily, might just be a matter of css prefixes or alternatives.

ascalamogna:
@danielfdsilva @dereklieu the only problem i'm seeing is the use of flexbox in the page body, which I believe we can remove, otherwise I think this is fine? Unless I'm missing something? I've been testing on IE 11 in Windows 10.

danielfdsilva:
@ascalamogna seems to be working well. merging

PR #89, opened 2018-01-08
Deployments Page Updates

ascalamogna:


danielfdsilva:
Fixes #95 
I think this page is ready to go. There are a couple of things that concern me with this page, namely the amount of queries needed to get all the data for the tables + map, but as per our discussion this is what is possible at the moment.

With the decision to go with a separate table for each deployment they became a bit poor. There's not much information on them, and the type of info needed was not listed anywhere so that has to be reviewed (cc @ascalamogna ). This probably has to do with being fake data, right?

Also to be reviewed is the map data and how the values are being calculated. The formula was not described anywhere so I'm not sure if I missed anything.

dereklieu:
@danielfdsilva I'm not crazy about the number of requests either, but the responses should be fairly small and the CDN should cache these well. Load times at least here in Malaysia aren't any worse than elsewhere on the site.

@ascalamogna could you take a look at the current iteration of this and make any adjustments to the text or style as you see fit? Also ongoing are changes you recommended to the API. It would be helpful to know what further changes are needed to this interface after the API has been updated.

ascalamogna:
@dereklieu @danielfdsilva 

I styled the deployments map. I think we should size the circles on the countries by number of deployments (excluding ERU's) so a country with 2 FACT would be larger than a country with 1 FACT. 

The tables should have the following headers. We either have these in the API already or they are documented in the ticket to be added.

ERU
Name (society), Country Deployed to, Type, Emergency, Number of Personnel

FACT
Start Date, Country Deployed to, Emergency 

RDRT/RIT
Start Date, Country Deployed to, Emergency 

HeOps
Start Date, Name, Country deployed to, Emergency

danielfdsilva:
@ascalamogna Correctly sizes the circles and changes the table information.
The emergencies all show up as N/A because there's no info returned about it in the response. cc @dereklieu 

Please have a look.

ascalamogna:
This looks correct to me.

guipizzini:
@ascalamogna any ETA of when this pull request will be merged?

PR #99, opened 2018-01-12
Feature/password retrieval

dereklieu:
Hook up password retrieval, password change views

### DIFF
--- a/app/assets/scripts/actions/index.js
+++ b/app/assets/scripts/actions/index.js
@@ -36,6 +36,11 @@ export function validateAndUpdatePassword (payload) {
   return postJSON('change_password', CHANGE_PASSWORD, payload);
 }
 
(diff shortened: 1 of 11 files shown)


danielfdsilva:
Looks good and frontend wise it works, merging!

Two notes:
- The link sent to the email to recover the password doesn't work, although that is not related to the frontend.
- I find it weird that after you change your password you're not immediately redirected but instead you get a message saying that you will be. Wouldn't it be better to redirect and saying "password successfully changed". What are your thoughts?

@dereklieu

dereklieu:
@danielfdsilva is there a good way to show the alert *after the redirect occurs? If so I think that would be the better option. 

On the email link: yeah that's not reflective of the development url at the moment.

danielfdsilva:
@dereklieu you can trigger the alert and redirect immediately. It will persist even if the triggering component is unmounted

PR #107, opened 2018-01-16
Hook up situation reports

dereklieu:
@danielfdsilva this hooks up situation reports.

Right now emergencies don't really have situation reports. I added one to this emergency though:

`/emergencies/3048`

You can also add more documents to this emergency using this [the admin](https://dsgocdnapi.azureedge.net/admin/api/event/3048/change/).

Note about the admin, you can select *either* to upload a file or supply a URL. This is to support files that are already hosted elsewhere. There's no good way to enforce this in the admin, but the front-end will use the uploaded file first if there is both an uploaded file and a supplied URL.

### DIFF
--- a/app/assets/scripts/actions/index.js
+++ b/app/assets/scripts/actions/index.js
@@ -122,6 +122,11 @@ export function getEventById (id) {
   return fetchJSON(`api/v1/event/${id}/`, GET_EVENT, withToken(), { id });
 }
 
+export const GET_SITREPS = 'GET_SITREPS';
+export function getSitrepsByEventId (id) {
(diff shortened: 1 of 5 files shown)


danielfdsilva:
@dereklieu Fixed some linting errors and fold structure.
The same situation report is being pulled for every emergency. Is that intentional?

dereklieu:
@danielfdsilva thanks! The same situation report being pulled was a bug in the API that I've now fixed. Reports should only be showing up now for the emergencies they are linked to.

danielfdsilva:
@dereklieu All good. merging!

PR #112, opened 2018-01-23
Guard against array out of bounds error

dereklieu:
This guards against an error I was getting where some time intervals existed for Drefs but not Appeals, which is not an impossible scenario.

For example, if it's the start of the month and a Dref has been filed, but not an Appeal, this would cause the prior view to throw.

@danielfdsilva let me know what you think of this fix.


dereklieu:
@danielfdsilva yep, it looks like this:

screen shot 2018-02-03 at 6 28 59 pm

The last month just shows as 0, which I think is fine. I will add a comment and merge.

PR #122, opened 2018-01-25
Hookup appeals docs

dereklieu:
Removes the appeals document placeholder and renders them

@danielfdsilva can you give this a look? You can check `/emergencies/6` for one that has appeal documents.


danielfdsilva:
Looks good. Merging

PR #149, opened 2018-01-30
Miscellaneous feedback items

dereklieu:
@ascalamogna could you take a look at the following items:

1. The "beta" message that's now in the header (feel free to move it if you think it needs it).
1. The full-screen deployments implementation.
1. Some fixes to the deployments page now that the structure has been changed.


ascalamogna:
@dereklieu can we add the ability to close the beta banner after it's seen? I put an x-mark in the note to allow this.

guipizzini:
@ascalamogna I like the idea of being able to close the beta banner. Thanks

PR #257, opened 2018-06-07
update country page layout

ascalamogna:


ascalamogna:
@dereklieu I'm including the country page layout here for reference:
[Country Page.pdf](https://github.com/IFRCGo/go-frontend/files/2081492/Country.Page.pdf)

ascalamogna:
@dereklieu to both pages we need to add the additional graphics and key figures modules. We also need to update the tables. 

I think both could benefit from having the sub navigation added.

PR #285, opened 2018-06-19
add export buttons

ascalamogna:
@dereklieu this has an export button for the map on country pages but really just place this anywhere you need a button before the `fold__title`

               <div className='fold__actions'>
                  <button className='button button--primary-bounded'>Export Map</button>
                </div>


dereklieu:
Maps can now export 💥

PR #437, opened 2018-09-11
Changing page style for test environment

szabozoltan69:
The requirement was: to change test environment colors so that testers should not miss where they are and do not type in test data to production environment. Now the environment variable is wired-in (to _variables.scss) and it should be changed, but temporarily it will be usable.


app/assets/styles/core/_base.scss
```diff
@@ -33,11 +33,17 @@ html {
...
-  background: tint($base-color, 98%);
-  color: $base-font-color;
+  @if $production_environment == true {
+    background: tint($base-color, 98%);
+    color: $base-font-color;
+    font-family: $base-font-family;
+  } @else {
+    background: #88AADD;
+    color: #0000FF;
+    font-family: "Cambria, Georgia, serif";
```
dereklieu: @szabozoltan69 I don't think you should change the font family under dev, a different color would be enough, no?

Also, I think it's better to control this from the javascript, where you have more knowledge of what environment something is in.

szabozoltan69: Ok, any other solution is welcome. I thought to change font family because it gives a basic style, "feeling", and the aim was to see at once that something is different here. But it was just an idea, we can work with colors only, if nicer.

PR #454, opened 2018-09-20
Fix env

dereklieu:
@szabozoltan69 this was showing the dev style in production.
Note that we use single quotes `'` and not double quotes `"`.


szabozoltan69:
Ok (single quote).
But I don't really understand why is better "=== staging" than "not production"...

dereklieu:
It turned out we were reading the wrong variable. This fixed that: https://github.com/IFRCGo/go-frontend/pull/456/files

Issue #364, opened 2018-07-23
Export CSVs with modified headers and columns


dereklieu:

LukeCaley:
We need to add sensible fields to the .csv export so this is understandable.  Revised field names in first row of the below: 
[2018-07-appeal.zip](https://github.com/IFRCGo/go-frontend/files/2223867/2018-07-appeal.zip)

dereklieu:
As of #371 CSVs save with a proper extension, and have their query limit removed. 

As it turns out, it's not simple to replace header fields. The payload from the API is already formatted as CSV, which to the browser is just text, and would require some expensive parsing and then re-formatting back to CSV in order to replace headers and fields. This would also negatively impact performance on whole-table queries, which is currently the default behavior.

The place to hook in this feature is probably at the API level, ie you could specify a list of columns and also a column name for each column to include in the resulting response. However, the API currently doesn't support this, and it's out of scope for phase 2.

I've changed the title of this ticket to reflect that. cc @LukeCaley

ElsaRaunioIFRC:
This issue has been partially solved. 

There are still some columns that I would revise, like this:

country society_name > national_society_name
code > appeal_code

Other than this, the column names are clear and understandable.

szabozoltan69:
Done. Tested with .../appeals/all?region=3

ElsaRaunioIFRC:
Let's go for -

szabozoltan69:
Done.

PR #472, opened 2018-10-04
Feature 2.1/issue 364

szabozoltan69:
Following API #364 changes with dtype reading


dereklieu:
Doesn't this rely on the API route being altered first, to send the dtype as text rather than as an id?

szabozoltan69:
This should be merged along with the feature_2.1/issue364 (missing underscore) branch in go-api.

PR #473, opened 2018-10-10
Replacements in CSV file headers

szabozoltan69:
Be the header more verbose and less technical hardcoded.

### DIFF
--- a/app/assets/scripts/components/export-button.js
+++ b/app/assets/scripts/components/export-button.js
@@ -19,7 +19,14 @@ class ExportButton extends React.Component {
 
   componentWillReceiveProps (newProps) {
     if (this.props.csv.fetching && !newProps.csv.fetching && !newProps.csv.error) {
-      const encodedUri = encodeURI('data:text/csv;charset=utf-8,' + newProps.csv.data);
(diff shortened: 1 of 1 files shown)


app/assets/scripts/components/export-button.js
```diff
@@ -19,7 +19,14 @@ class ExportButton extends React.Component {
 
   componentWillReceiveProps (newProps) {
     if (this.props.csv.fetching && !newProps.csv.fetching && !newProps.csv.error) {
-      const encodedUri = encodeURI('data:text/csv;charset=utf-8,' + newProps.csv.data);
+      var thisdata = newProps.csv.data.replace(/\n/,'¤𝛼¤ß¤');
+      var firstLine = thisdata.split('¤𝛼¤ß¤')[0];
+      var moreLines = thisdata.split('¤𝛼¤ß¤')[1];
```
dereklieu: This creates two potentially very large arrays when you could simply create one. 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice

szabozoltan69:
Yes, this is true. I (or you) should solve it via array slice instead of duplication.

PR #547, opened 2018-12-04
Finishing username notification

szabozoltan69:
Adding email reducer


szabozoltan69:
@dereklieu I removed the error checking (like https://github.com/IFRCGo/go-frontend/blob/develop/app/assets/scripts/views/recover-account.js#L38 ) from https://github.com/IFRCGo/go-frontend/blob/develop/app/assets/scripts/views/recover-username.js#L33 – this way a possible attacker can not test existing email addresses. The error message comes from the API, but the frontend does not show it.

Issue #424, opened 2018-08-23
Sources for summary text on emergency page


guipizzini:
## Issue

We do not have any section to add the "source" in the Django framework but yet the source will appear automatically into the text once pushed to the front end. As well, the formatting and layout of how the source appears is not clean and well structured (should be separate from the text, with a smaller font and in bold highlighting as well the date of the piece of information).

ElsaRaunioIFRC:
@szabozoltan69 How is the issue with source and formatting? Without an access to the dev admin, I can not do any testing.

The current field reports don't seem to have any paragraphs - I assume it is still not possible to divide the overview into paragraphs? Would it be difficult to introduce this option?

szabozoltan69:
@ElsaRaunioIFRC I have no experience about formatting GO input. It can be time-consuming to explore it. I suggest to forward this to phase 3.

szabozoltan69:
In emergency.js and field-report.js we should use dangerouslySetInnerHTML instead of plain text according to GO III Req21. (Instead of emergency.js the expandable.js was a fine approach, but in case of the "Show more" link I failed to use a simple "span" instead of React.Fragment.)
In expandable.js we could use an "if (text.length > limit)" condition for displaying (or not) the "Show more" link.

In field-report.js this is the relevant row ( https://github.com/IFRCGo/go-frontend/blob/develop/app/assets/scripts/views/field-report.js#L202 )
```
<DisplaySection title='Description' inner={get(data, 'description', false)} />
```

nanometrenat:
Original issue was re ACAPS API - which is no longer available so closing this ticket.

nanometrenat:
The formatting aspect has been spun off into new ticket #931

PR #550, opened 2018-12-06
Format and style emergency sources

dereklieu:
To address https://github.com/IFRCGo/go-frontend/issues/424

screen shot 2018-12-06 at 3 18 33 pm

screen shot 2018-12-06 at 3 18 38 pm


dereklieu:
If the source refers to a field report, it will automatically pull in the summary as the link, as well as the updated at timestamp (or created at timestamp if updated at does not exist). Sources that refer to the emergency will show the emergency name, but will not have a link since it would link to the page the user is currently on.

PR #585, opened 2019-03-05
Update field-report-constants.js

ElsaRaunioIFRC:
After a discussion with the IFRC Governing Board, we agreed to change "The former Yugoslav Republic of Macedonia" to "North Macedonia, Republic of".

### DIFF
--- a/app/assets/scripts/utils/field-report-constants.js
+++ b/app/assets/scripts/utils/field-report-constants.js
@@ -126,7 +126,7 @@ export const countries = [
   {value: '105', label: 'Liechtenstein', iso: 'li'},
   {value: '106', label: 'Lithuania', iso: 'lt'},
   {value: '107', label: 'Luxembourg', iso: 'lu'},
-  {value: '108', label: 'The former Yugoslav Republic of Macedonia', iso: 'mk'},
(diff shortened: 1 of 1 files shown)


mankamolnar:
Hi! I`ve approved and merged your PR. The next release probably will happen next monday/thuesday (2019 March 11-12).

ElsaRaunioIFRC:
Thank you for the collaboration, @mankamolnar

mankamolnar:
@ElsaRaunioIFRC Sorry just one question. This way it will show 'North Macedonia, Republic of'. Are you sure it`s ok? that 'Republic of' looks weird at the end on it`s own.

ElsaRaunioIFRC:
@mankamolnar I agree that "Republic of" at the end looks a bit weird - however, the IFRC applies this logic to some country names. Other examples include "Iran, Islamic Republic of", "Tanzania, United Republic of" and "Gambia, The". The [UN OCHA](https://fts.unocha.org/countries/overview) also uses the form "North Macedonia, Republic of".

In the future, we may use simply "North Macedonia" - however, we need an approval for this from the country cluster office and the governing board.

mankamolnar:
@ElsaRaunioIFRC Ok, I just wanted to be sure that everything is right. Thank you

Issue #640, opened 2019-06-26
3w: Form to capture project details


batpad:
We need to implement a form on the frontend for users to enter details of Projects, that will be aggregated on the Who-What-Where frontend on the Country page.

The design for the form is here: https://projects.invisionapp.com/share/P9SA110GT4S#/screens/369905382

Backend tickets to implement [models](https://github.com/IFRCGo/go-api/issues/443) and [API](https://github.com/IFRCGo/go-api/issues/444).

This form will post to the API we create on the backend and save `Projects`.

Question on the design: currently, there is a button next to the map on the country page that says `Edit` that takes you to this form. I'm not sure how that's supposed to work, or if that button should say `Add` ? If it is meant to be `Edit`, which Project should they be editing, since there will be many projects? I will clarify this with the design team.

cc @mankamolnar

batpad:
Capturing some changes to the design based on discussion with @guipizzini : 

 - Add a field called `Type of Operation` (radio?) with options `Long-term operation` and `Emergency operation`
 - Move `Programme Type` field to next to or below `Type of Operation` field
 - Based on selection of `Type of Operation` and `Programme Type` fields:
   - if Type of Operation is `Emergency Operation` and `Programme Type` is `Multilateral`, show `Current Operation` dropdown. Else, do not show. @guipizzini to confirm if `Current Operation` is to show or not for `Emergency Operation` that is of Type `Bilateral`.
 - Split `Sector of Activity` into two fields:
   - `Primary Sector of Activity` which will allow a user to select only a single sector.
   - `Secondary Sectors of Activity` which is a multi-select, currently will have same options as `Primary Sector of Activity`, but user can select multiple. Backend ticket for this: https://github.com/IFRCGo/go-api/issues/514

@frozenhelium @guipizzini - please let me know if I missed capturing anything.

frozenhelium:
More changes after discussing with @guipizzini and @LukeCaley

batpad:
@LukeCaley both @frozenhelium and my strong recommendation here would be to go for option 1 for this - when the user updated the `status` to `complete`, show the 2 numeric fields for `People Targeted` and `People reached` and make these fields required if status is complete.

This will guarantee that we can validate the data and make sure that there cannot be Projects where the status is complete and do not have this data filled out. Overall, this seems to me like it's more intuitive for the user, as well as easier to implement.

[...]

PR #650, opened 2019-07-01
[wip] Form to add / update Projects (part of 3w)

batpad:
Towards https://github.com/IFRCGo/go-frontend/issues/640

This starts sketching out the form on the frontend to add and edit Projects data.

So far, it adds all the actions and reducers and mappings to the state. I'm struggling a bit with figuring out laying out the various form components as per the design. I'll probably go ahead and just use the form components available and make things functional, and move onto the layout as a second step.

@mankamolnar I followed the structure of the `field-report-form` where possible - I think this all makes sense and I should be able to go ahead with adding the form components, validators, etc. Let me know if you have any comments / feedback.

### DIFF
--- a/app/assets/scripts/actions/index.js
+++ b/app/assets/scripts/actions/index.js
@@ -439,3 +439,18 @@ export const EDIT_PER_DOCUMENT = 'EDIT_PER_DOCUMENT';
 export function editPerDocument (data) {
   return postJSON('editperform', EDIT_PER_DOCUMENT, data);
 }
+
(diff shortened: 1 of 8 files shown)


batpad:
Closing as we are not implementing in this phase.

batpad:
Closing this PR because it is out of scope for this phase of work. We should definitely refer to this PR and try and pick up where it leaves off when we start implementing the form for 3w projects. #640 .

PR #689, opened 2019-07-22
Field report changes

batpad:
This implements the field report changes as per the new design:

It reorganizes the fields into 4 steps as per the design, and matches the layout to the new design.

We should ideally merge this as is after a bit of testing.

What is remaining:

 - Add new fields: Start Date + Additional Contact field
 - Display fields and labels conditionally based on Field Report status

^ I would rather work on these in a separate pull request.

cc @mankamolnar


batpad:
Just fixing the lint issues ..

batpad:
@szabozoltan69 this is a very big PR that makes many big changes to the field report form. If you have the time, can I ask you to pull this branch and just test the field report functionality that it seems to work okay according to you? Then we can merge to `develop` and ask others to also test while I work on the more advanced changes.

szabozoltan69:
Thanks for the "dont throw error if both title and Event link are filled" fix, @batpad .
We could now merge if you agree.

PR #714, opened 2019-07-24
Feature/ui prep styling

batpad:


batpad:
Cleans up and makes consistent some styling on the Country -> Preparedness section

cc @mankamolnar

PR #717, opened 2019-07-24
Field report fields: add Other source and start_date

batpad:
Adds start_date and "Other" source.

It would be nice to give this a bit of a thorough test before merging. @szabozoltan69 if you are able to test this and it works, please feel free to merge, otherwise I will give it a look again in the morning.


szabozoltan69:
A few minutes, I am on it...

PR #720, opened 2019-07-25
Adds field for Contact - Phone to Field Report

batpad:
cc @szabozoltan69 

If you are able to do a quick check and merge, hopefully @mmusori can test further -- this adds the "Phone Number" field to the Contact info on the Field Report.

I will make the other changes mentioned in the document now / bringing text in line with Invision, etc.


szabozoltan69:
Ok, but backend side I have to deploy the merged master branch; but I shall be ready in minutes.

PR #736, opened 2019-07-30
Creates tab component for countries page

necoline:
The tabs use react-tabs to replace the sticky page nav. 

The page loads with the first tab displayed unless otherwise specified in the url.

All tabs are displayed. The conditional display of content was moved from the menu to the content panel. If there is no data, an panel will display a message of 'coming soon'.

This error panel is held in a separate component with other common components

Since this data is not being changed, the tab details were moved outside of the class and stored as an enum. 

A separate scss file was created to style the tabs.


szabozoltan69:
Hi Guys,
 I made the "tabs" branch mergeable with develop. There were some conflicts, so I could make some mistakes, mainly concerning the countries.js script, which were some news from non-tabs side also. Check it please.

PR #747, opened 2019-08-02
Account for no regions

danielfdsilva:


szabozoltan69:
(Otherwise region == 0 is for African region, which is a strange thing – but this is the situation.)

Issue #774, opened 2019-08-23
GO 3.2 - Regional page redesign


LukeCaley:
Issue relates to implementing a redesign for the regional page following approved wireframes here: https://projects.invisionapp.com/share/K8R901UBXN4#/screens/360977415

General organisation of the page echoes the home page
- Cards with key figures
- Highlighted operations (that the ROs can set themselves)
- Country list (filter to be alphabetical or IFRC-supported operations)
- Active operation 

as well as further down the line charts to echo data around appeals and DREFs by year and by month for the region.

batpad:
Key Figures: These should all be exactly the same as the home-page, except filtered by region.

batpad:
Highlighted Operations: filter highlighted by region (@szabozoltan69 may need your help in this a bit)

batpad:
Country List: Change "Active Operations" To "IFRC Supported Operations" in operations count under the countries.

batpad:
Country Filter: Filter countries down to only countries that have active operations (not priority).

batpad:
Graphs: Should show exact data as home-page, only filtered by region.

guipizzini:
@batpad let us know if we can help with this or simply if you have feedback on progress. Thanks!

batpad:
> Active operation

@LukeCaley can you clarify what this refers to? Is this just consolidating the Active Operations table + map to look similar to the home-page layout? Or is this something else?

necoline:
We have concluded Release 3.2 and are currently working on 4.1. As such, all of the remaining UI/UX enhancements will reference this upcoming release instead of tracking on this issue.
For reference, below are all of the PRs related to the issue in this last release:

Adds date filter to home and region map: #838 
Updates region map to match home page: #842
Lists countries in alphabetical order: #792

batpad:
@teklal can I ask you to test and verify this all works as expected? cc @guipizzini

LukeCaley:
User test accepted as per @necoline list of components above (Adds date filter to home and region map: #838; Updates region map to match home page: #842; Lists countries in alphabetical order: #792).  Proceed to deploy

PR #792, opened 2019-09-02
fix display of alphabetical country list on region page, toward #774

batpad:


batpad:
Display countries as a 3 column list, alphabetically top to down rather than left to right. Also add "headings" with the first letter of the alphabet for each section.

cc @thenav56 @frozenhelium

Issue #785, opened 2019-08-29
Login button inconsistency


teklal:
When a user is not logged into GO there is a login button on the top-right corner of the front page (as there should be).
However, when this same page is viewed with a narrow window width, the respective 'button' on the dropdown menu says 'Logout' although it should say 'Login'

Below screenshots of the same window in these two different window width scenarios.

![gologin](https://user-images.githubusercontent.com/52411237/63928286-9b67a980-ca4f-11e9-8ad0-efaa1e5eae8d.PNG)

![gologout](https://user-images.githubusercontent.com/52411237/63928287-9b67a980-ca4f-11e9-8494-42d5f677b974.PNG)

batpad:
@teklal thank you for the catch.

This is a pretty serious issue.

The problem here is that we have two different header components based on whether the user is on mobile or desktop.

For mobiles, this header component is used: https://github.com/IFRCGo/go-frontend/blob/develop/app/assets/scripts/components/mobile-header.js

One can see the account links are hard-coded, as the current user login data is not being passed down to the mobile header: https://github.com/IFRCGo/go-frontend/blob/develop/app/assets/scripts/components/mobile-header.js#L118

So we need to ensure the users current login details are correctly passed down to the mobile header, and then write the conditional code to handle displaying Login / Logout, etc. based on the user's login state.

@necoline let's talk about this, and see if it might be something you're able to dig into.

necoline:
Sounds like a plan.

guipizzini:
@batpad and @necoline if you can share any update or feedback on your discussion that would be very useful. thanks friends

batpad:
This was fixed in #805 and is ready for testing @teklal @guipizzini .

Thanks @necoline for the fix!

PR #805, opened 2019-09-05
Fix/785 login on mobile menu

necoline:
#785 Adding user-menu component to the mobile menu

Previously, there was just a hard coded logout button
I imported user-menu and then updated user menu to only display the logout button if the view in on mobile rather than display the dropdown menu that is used in the non-mobile layout.

I also made a small fix to the mobile menu background. 
The height was set to 100vh so the menu items extending beyond that had no background so it was sitting on top of page content

### DIFF
--- a/app/assets/scripts/components/connected/user-menu.js
+++ b/app/assets/scripts/components/connected/user-menu.js
@@ -9,6 +9,28 @@ import { logoutUser } from '../../actions';
 import Dropdown from '../dropdown';
 
 class UserMenu extends React.Component {
+  constructor () {
+    super();
+    this.state = {
+      width: window.innerWidth
+    };
+  }
+
+  // Adds a listener for window size to determine style for menu content
+  componentWillMount () {
+    window.addEventListener('resize', this.handleWindowSizeChange);
+  }
+
+  // Ensures the listener is removed when the component is no longer mounted
+  componentWillUnmount () {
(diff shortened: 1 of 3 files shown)


app/assets/scripts/components/connected/user-menu.js
```diff
@@ -9,14 +9,40 @@ import { logoutUser } from '../../actions';
...
+  constructor () {
+    super();
+    this.state = {
+      width: window.innerWidth
+    };
+  }
+
+  // Adds a listener for window size to determine style for menu content
+  componentWillMount () {
+    window.addEventListener('resize', this.handleWindowSizeChange);
```
batpad: I am slightly concerned that this might get called a lot as the user is resizing the window. While testing though and trying to resize a lot, I did not notice any weird behaviour, so I'm going to go ahead and merge. We may want to add something like a `debounce` to the function in the future: https://lodash.com/docs/4.17.15#debounce

batpad:
This seems to work really well and the code looks good to me - I really like the idea of putting the same `user-menu` component into the mobile header - it's something I some-how had not considered.

app/assets/scripts/components/connected/user-menu.js
necoline: That's a fair concern. It's not ideally set up for users to resize the window. I don't imagine that users will resize fairly often, but I will try to re-jig this in a new pr so that it won't come up as an issue.

Issue #773, opened 2019-08-22
Response Documents organisation through cards


LukeCaley:
In major operations there are often dozens of Response Documents and it can be tricky to find the one you are looking for.  Having 'empty' cards for expected documents might also suggest to users to upload these 'missing' documents.

This was quite a big issue in the Mozambique operation for example: https://go.ifrc.org/emergencies/3469#details (need to scroll down to the bottom of the page).

There are wireframes here: https://projects.invisionapp.com/share/K8R901UBXN4#/screens/361592718 which also suggest a date range filter and search function, both of which would be nice-to-haves but are not absolutely essential at this stage.  It would however be good to be able to directly upload a missing doc type, redirecting the user to the relevant place in the back-end., through a simple upload button on the card itself.

Related to #396 and #413

batpad:
For Response Documents, use Situation Report Types defined in the backend to show the 6 groupings of report types.

LukeCaley:
@batpad as discussed, more info on the Response document types, and text for callouts, for the front end:

1. Situation reports - "IFRC, National Society, and other Emergency Response Unit and technical sitreps"
2. Maps / infographics - "Static maps and graphics created by IFRC and Movement partners"
3. Needs assessment reports - "IFRC, National Society and Movement Partners"
4. Surge documents - "Staff health, security, ToRs, and other surge related docs")
5. Mob table and logistics - "Mob tables, shipping details and other logistics docs"
6. Secondary data links - "Key information sources for situational awareness" (this should be a "link only" card - with no document upload)

@batpad If at all possible, would we be able to implement a 'pin' doc function so that we could identify the key documents within each type and ensure that they are seen first?

LukeCaley:
One to one mapping of revised Response Documents categorisation:

1. Situation reports - contains 'Situation Reports' and 'ERU reports'
2. Maps / infographics - 'Maps'
3. Needs assessment reports - 'Information products' 
4. Surge documents - 'Key Surge Documents'
5. Mob table and logistics - 'Mobilisation Tables'
6. Secondary data links - no current document type in the d/b

Note to self @guipizzini @mmusori @tekla to review the implications on previously categorised 'Information products' which are not necessarily 'Needs assessment reports'.

LukeCaley:
In terms of ordering:

One to one mapping of revised Response Documents categorisation:

First row:
1. Situation reports 
2. Surge documents 
3. Mob table and logistics
Second row
4. Maps / infographics 
[...]

PR #810, opened 2019-09-09
Fix/773 order response docs

necoline:
#773 Adds a sort function on the response doc util to provide records in order by title


batpad:
@necoline I just changed the ordering to use the ids instead of titles, as they are likelier to be stable. Also added a "sort by date" so that the response documents should now be sorted by `created_at` timestamp descending.

Would be great to get a quick review.

PR #811, opened 2019-09-10
hide language selection drop-down on PER forms, fixes #808

### LINKED ISSUE #808: PER language forms to be hidden
## Issue

PER forms were translated by the PER Team. However, due to last minute fixes on the forms and due to the fact the rendering of the front end of GO is in English only - we had agreed to keep the functionality so we can easily make the translated forms available - but to **hide** the forms from the user's front end for now.

## Steps to reproduce

Check the screenshot - the highlighted dropdown menu - needs to be hidden.

## Expected behavior

Forms available only in english for PER.
Screenshot 2019-09-08 at 12 14 10

batpad:
I'm having a bit of a hard time figuring out testing this, see https://github.com/IFRCGo/go-frontend/issues/808#issuecomment-529891501

But, from what I can tell, this *should* work.

@necoline can I ask you to give this a quick look and let me know if this seems sane? We want to (temporarily) hide the language selector on these forms as per #808 .

I will ask the team to test thoroughly once we merge this to develop to check that this does indeed work.

### DIFF
--- a/app/assets/scripts/components/per-forms/per-form-component.js
+++ b/app/assets/scripts/components/per-forms/per-form-component.js
@@ -8,6 +8,13 @@ import RequestFactory from './factory/request-factory';
 const requestFactory = new RequestFactory();
 
 const renderLanguageSelectDropdown = (props) => {
+  /*
+  The language selection code is commented out as per
+  https://github.com/IFRCGo/go-frontend/issues/808
+
+  Intentionally leaving this commented out because we will want to
+  bring this functionality back.
+
   return (<div>
     <span style={{fontWeight: 'bold'}}>Form language:</span>&nbsp;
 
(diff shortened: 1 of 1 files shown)


batpad:
Tested this locally and it seems to work well.

To test:

 - Go to Account -> PER Forms, and then click on one of the Pills with the form types to see a form.
 - With this branch, the forms should no longer show a language selector, and just default to english.

necoline:
Great. Let's leave it as is. This way makes it easy to undo later

batpad:
> This works. I would ideally go for a disable feature especially if there is cause to think that we will need to do this again in the future. It would also signal to the user that this is a feature that exists and will be enabled in the future.

I think this is a good suggestion. I'm going to merge this, but comment on that ticket asking if they would be okay with displaying the language drop-down but having it disabled.

Issue #823, opened 2019-09-13
Disable fullscreen and map filtering functionalities for Deployments page


teklal:
The fullscreen mode on the current 'Deployments' page on GO doesn't work well at the moment so it would be better to disable this functionality for now. Below a screenshot of what the fullscreen view looks like at the moment.

![image](https://user-images.githubusercontent.com/52411237/64862335-833c8080-d632-11e9-816e-aafceb21c854.png)

The filtering functionality doesn't fully work either, so this should be disabled as well.
There are currently FACT and RDRT deployments and ERUs in The Bahamas, but the map doesn't show the ERUs when that filter is chosen.

![image](https://user-images.githubusercontent.com/52411237/64863135-74ef6400-d634-11e9-9aa0-b789ab4067ac.png)

![image](https://user-images.githubusercontent.com/52411237/64863154-7fa9f900-d634-11e9-8521-952fd37d0a5b.png)

![image](https://user-images.githubusercontent.com/52411237/64863186-905a6f00-d634-11e9-9e59-4fdda4368714.png)

@szabozoltan69

szabozoltan69:
Done via https://github.com/IFRCGo/go-frontend/pull/824

szabozoltan69:
The functionalities should be fixed, so I reopen this ticket.

teklal:
Closing as Deployments/Surge is going to undergo database changes in the coming months.

Issue #822, opened 2019-09-12
Field Report labels


mmusori:
The data entry side of the Field report has the right terminology “Actions taken by any other RCRC Movement actors ” but in the output report that is then sent out there is a different label “Actions taken by PNS Red Cross”. Case in point is the  Chile Field Report  from 12/09/2019.

szabozoltan69:
Done via https://github.com/IFRCGo/go-frontend/pull/824

PR #824, opened 2019-09-13
Disable fullscreen and map filtering functionalities for Deployments …

szabozoltan69:
…page #823, Field Report labels #822


teklal:
Thanks @szabozoltan69 !

PR #827, opened 2019-09-17
Add 3W tab and project form in countries page

frozenhelium:
Expected behaviour for project form:
* Shows error below each required field if not filled before submission
* Failure to submission (due to network error or server error) is not shown currently
* Form automatically closes on successful submission

Currently missing in project form:
* Mechanism to select regional project
* Disabling country input according to the user permission

Currently missing in 3W page
* Link for the event (Current operation field)
* Edit project button
* Hiding add project button according to the user permission

Also the mechanism to show global loading in the countries page is updated.

P.S. I've used a `new` directory to keep track of the new components (they still need some finishing touch), which I'll move to appropriate places accordingly in the future.


app/assets/scripts/views/ThreeW/filter.js
```diff
@@ -0,0 +1,101 @@
+'use strict';
+import React from 'react';
+import { PropTypes } from 'prop-types';
+import Faram from '@togglecorp/faram';
+
+import Select from '../../components/form-elements/new-components/select';
```
batpad: Can we avoid things like `new-components` ? Is there a better way to separate things out? `new-components` will no longer make sense down the line. Ideally, would we replace all the current Selects with the new Select and then we no longer need this? Am just a little bit worried about this getting confusing - it's ok for now if we are transitioning and don't want to make too many changes in a single PR, but in this case, can we add a quick README or so to the `new-components` folder to just explain why it exists, and make a note that ideally sometime down the line we'd consolidate the Selects on the site to this and then no longer need the `new-components` folder?

app/assets/scripts/components/new/select.js
```diff
@@ -0,0 +1,56 @@
+'use strict';
+import React from 'react';
+import PropTypes from 'prop-types';
+import Select from 'react-select';
+import { _cs } from '@togglecorp/fujs';
```
batpad: @frozenhelium we are already using a library called `classnames` in the project that seems to do something very similar. You can see an example of usage here: https://github.com/IFRCGo/go-frontend/blob/develop/app/assets/scripts/components/fold.js#L22

It's used in many components and it would be nice to stay consistent. I'd rather stick with the existing library for now -- if there's a good reason to use the new library, let's make a ticket to move over all the existing usage of the `classnames` library to use the `_cs` method from `@togglecorp/fujs`. It would be really nice to not have multiple ways of doing this across the code-base :-)

app/assets/scripts/components/new/donut-chart.js
```diff
@@ -0,0 +1,204 @@
+'use strict';
```
batpad: Can the `donut-chart` component just go in `scripts/components/` - I don't see a reason for this to be under `new` ?

app/assets/scripts/components/new/select.js
frozenhelium: Sure, I'll check if they have same set of features and take action accordingly

app/assets/scripts/components/new/donut-chart.js
frozenhelium: I'm using the `new` directory to keep track of the things I added, I'll move these components appropriately later on

app/assets/scripts/views/ThreeW/filter.js
frozenhelium: This is temporary and I'll think of something better

batpad:
@frozenhelium any idea why CI is failing? Seeing:

```
yarn run v1.19.1
$ eslint app/assets/scripts/ --ext .js
Cannot read config file: /root/repo/.eslintrc
Error: missed comma between flow collection entries at line 13, column 18:
        "ecmaVersion": 6,
                     ^
YAMLException: Cannot read config file: /root/repo/.eslintrc
Error: missed comma between flow collection entries at line 13, column 18:
        "ecmaVersion": 6,
                     ^
    at generateError (/root/repo/node_modules/js-yaml/lib/js-yaml/loader.js:167:10)
    at throwError (/root/repo/node_modules/js-yaml/lib/js-yaml/loader.js:173:9)
    at readFlowCollection (/root/repo/node_modules/js-yaml/lib/js-yaml/loader.js:718:7)
    at composeNode (/root/repo/node_modules/js-yaml/lib/js-yaml/loader.js:1360:11)
    at readFlowCollection (/root/repo/node_modules/js-yaml/lib/js-yaml/loader.js:746:7)
    at composeNode (/root/repo/node_modules/js-yaml/lib/js-yaml/loader.js:1360:11)
    at readBlockMapping (/root/repo/node_modules/js-yaml/lib/js-yaml/loader.js:1036:16)
    at composeNode (/root/repo/node_modules/js-yaml/lib/js-yaml/loader.js:1359:12)
    at readDocument (/root/repo/node_modules/js-yaml/lib/js-yaml/loader.js:1519:3)
    at loadDocuments (/root/repo/node_modules/js-yaml/lib/js-yaml/loader.js:1575:5)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
Exited with code 1
```

Do you have an idea of what this is / how to fix it?

app/assets/scripts/utils/field-report-constants.js
```diff
@@ -29,8 +30,7 @@ export const visibility = [
   }
 ];
 
-export const countries = [
-  {value: '-- Country --', label: ''},
+export const countryList = [
```
batpad: @frozenhelium I'm a bit worried about this breaking other places in the app that use the `countries` list?

frozenhelium: it won't, the `countries` variable is same as before. I've just separated out the list of countries to `countryList` variable and made `countries` to extend `countryList`

frozenhelium:
> @frozenhelium any idea why CI is failing? Seeing:
> 
> ```
> yarn run v1.19.1
> $ eslint app/assets/scripts/ --ext .js
> Cannot read config file: /root/repo/.eslintrc
> Error: missed comma between flow collection entries at line 13, column 18:
>         "ecmaVersion": 6,
>                      ^
> YAMLException: Cannot read config file: /root/repo/.eslintrc
> Error: missed comma between flow collection entries at line 13, column 18:
>         "ecmaVersion": 6,
>                      ^
>     at generateError (/root/repo/node_modules/js-yaml/lib/js-yaml/loader.js:167:10)
>     at throwError (/root/repo/node_modules/js-yaml/lib/js-yaml/loader.js:173:9)
>     at readFlowCollection (/root/repo/node_modules/js-yaml/lib/js-yaml/loader.js:718:7)
>     at composeNode (/root/repo/node_modules/js-yaml/lib/js-yaml/loader.js:1360:11)
>     at readFlowCollection (/root/repo/node_modules/js-yaml/lib/js-yaml/loader.js:746:7)
>     at composeNode (/root/repo/node_modules/js-yaml/lib/js-yaml/loader.js:1360:11)
>     at readBlockMapping (/root/repo/node_modules/js-yaml/lib/js-yaml/loader.js:1036:16)
>     at composeNode (/root/repo/node_modules/js-yaml/lib/js-yaml/loader.js:1359:12)
>     at readDocument (/root/repo/node_modules/js-yaml/lib/js-yaml/loader.js:1519:3)
>     at loadDocuments (/root/repo/node_modules/js-yaml/lib/js-yaml/loader.js:1575:5)
> error Command failed with exit code 1.
> info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
> Exited with code 1
> ```
> 
> Do you have an idea of what this is / how to fix it?

I'll check it out

frozenhelium:
Fixed CI

app/assets/scripts/actions/index.js
```diff
@@ -213,6 +236,9 @@ export function getEruOwners () {
 
 export const GET_DISTRICTS = 'GET_DISTRICTS';
```
batpad: Leaving a note for myself here to check conventions for action names and see if this could be better.

app/assets/scripts/views/countries.js
```diff
@@ -659,13 +736,70 @@ class AdminArea extends SFPComponent {
     );
   }
 
+  syncBodyOverflow = (shouldOverflow) => {
+    if (shouldOverflow) {
+      document.getElementsByTagName('html')[0].style.overflow = 'hidden';
```
batpad: @frozenhelium is it possible to say a bit why this was needed / link to the CSS that made this required to do? Ideally we should never need to set the `overflow` of the top level HTML element to `hidden`, is there a better way we could be doing this?

app/assets/scripts/utils/field-report-constants.js
batpad: Missed that. Looks good :+1:

app/assets/scripts/views/countries.js
frozenhelium: It was required because the project form is overlaid on top of countries page, and the overflow from countries page was still visible.

frozenhelium: This function should have been memoized

frozenhelium: The better way would be to implement overflow in the country page itself and hide the overflow (if still required) from the css when the project form is shown. This can be achieved as

batpad: Ah! Let me try looking at this a bit. If I understand correctly, this is to be able to keep the overlay on top of the country page and prevent the page from scrolling? Hmm.. and `position:fixed` on the overlay does not work because it can get taller than the height of the screen?

Let's chat about this tomorrow and let me try and play around with it, but, right, I can see this being a bit tricky.

batpad:
@frozenhelium this looks really good! I'll do a bit more testing tomorrow, but by and large this is looking great!

My only slightly big request would be please move components out of `new/`. Here is what I propose:

 - For components that don't have existing similar components in the GO code-base, just add them to the `components/` (or whichever is the appropriate folder for each) folder.
 - For components that do have similar components in the existing code-base:
   - Create a ticket about the new component and the existing component (it can be a stub ticket for now), so we have a clear record of what components we need to rationalize and can use that ticket for discussion.
   - Link to this ticket in comments in the new component js file
   - Either rename the new components to `newComponentName>.js` and put them in the appropriate component folder, or leave them in `new/` and ideally add a quick `README.md` in `new/` linking to the tickets created above.

It is **really** important that we do not carry this separation into `new/` components forward and we should resolve as much as we can before merging. Happy to chat about how we can do this in a way that is least disruptive.

For components that are actually new and don't exist, I see no reason to leave them in `new/` - the code looks good to me and can always be improved down the line. For existing components where we are creating similar components, it would be nice to have this stand out and be bold and clear and documented, so we can find time to resolve it soon.

Looking forward to playing with this more tomorrow, fantastic work!

frozenhelium:
> @frozenhelium this looks really good! I'll do a bit more testing tomorrow, but by and large this is looking great!
> 
> My only slightly big request would be please move components out of `new/`. Here is what I propose:
> 
> * For components that don't have existing similar components in the GO code-base, just add them to the `components/` (or whichever is the appropriate folder for each) folder.
> * For components that do have similar components in the existing code-base:
>   
>   * Create a ticket about the new component and the existing component (it can be a stub ticket for now), so we have a clear record of what components we need to rationalize and can use that ticket for discussion.
>   * Link to this ticket in comments in the new component js file
>   * Either rename the new components to `newComponentName>.js` and put them in the appropriate component folder, or leave them in `new/` and ideally add a quick `README.md` in `new/` linking to the tickets created above.
> 
> It is **really** important that we do not carry this separation into `new/` components forward and we should resolve as much as we can before merging. Happy to chat about how we can do this in a way that is least disruptive.
> 
> For components that are actually new and don't exist, I see no reason to leave them in `new/` - the code looks good to me and can always be improved down the line. For existing components where we are creating similar components, it would be nice to have this stand out and be bold and clear and documented, so we can find time to resolve it soon.
> 
> Looking forward to playing with this more tomorrow, fantastic work!

I've moved the components from the `new/`  appropriately. All input components are put into `form-elements/` and remaining to `components/`. I'll create the ticket about them along with few more tickets regarding the refactors and other improvements in the system

Issue #828, opened 2019-09-26
Enable date filter on map (home + region pages)


batpad:
In addition to the `All Appeal Types` and `All Emergency Types` filters above the map, it should include a date filter, similar to the date filter on the table below.

There are two options here for the implementation:

 - Simpler: Add a start date and end date filter to the map filters, similar as the table. They operate only on the map (independently from the table filters) and filter the data on the map to only the data within the filtered date range. This will take a bit of looking into the data structure for the data on the map, and applying this filter to the GeoJSON.

 - Complex: Refactor the filters completely, and make the filters for the data on the map and in the table a common set of filters. In addition to being a considerable UI lift, this does introduce a lot of complexity since the data on the map and the data in the table are not the exact same data, and filtering through each with a common set of parameters could introduce weird edge-cases.

I'd vote for just adding a date filter to the map filters, but it would be good to make sure we're all on the same page with this implementation.

cc @LukeCaley @guipizzini @necoline

necoline:
Maybe we could start by throwing all of the filters into a common utility. If there is any overlap in logic, we can apply that to the maps and tables in the different views, otherwise we can breakout methods to deal with the unique data sets where needed. Would that be inline with what you are thinking @batpad ?

batpad:
>  Would that be inline with what you are thinking @batpad ?

@necoline I'm a bit worried about throwing all the filters into a common utility - it seems like the filtering is just so different between the table and the map, that it maybe more trouble than it's worth to have a common interface. I also think the design and interaction is not so obvious / clear if we want to "merge" the filters to have a common filter interface for the map and the table.

I would definitely vote for, to start with, just implementing the date filter on the map as a separate thing, separate from the filters on the table below. I think the work to re-imagine the filters to be common between the table and the map is probably a bigger piece, and if we have these filters implemented separately, it maybe easier to think about what parts are common and can be abstracted,  and also think about the user experience more carefully if we take the filters out of the table headings and need to represent them along-with the map filters.

Does that make sense?

necoline:
Per our conversation earlier, I reviewed the table and map more closely to see where the data is coming from. 

The data being used to populate the table and the map are the same however, the logic to manipulate the table and the map are very different. Some of this is necessary, other bits can be parsed out and used across both map and table if we want to. 

[...]

PR #838, opened 2019-10-08
map date filter

necoline:
#828 

This still needs a lot of work 
- styling
- removing duplicate time-date-header components


batpad:
@necoline this is looking really good to me. I can also confirm that the functionality seems to work well.

I do imagine the styling will be a little bit of a challenge - a bit curious to see how you are going to approach that :-)

After the date filter for the map is implemented, we can probably ticket out where things stand with the filters and why some are implemented differently and what longer term plans for the filters are, from a user / UI perspective as well as a code / refactoring perspective (more on this in #828 ).

This approach to harmonizing the date filter into a common component looks really solid - I do imagine the styling and little things will be a bunch of work - but I do think it's an approach we can re-use really well. Let me know if I can help or if it helps to talk through any of this.

necoline:
I don't have a clear roadmap for the styling. I am going to put some time into this today. My initial plan is to identify the component specific styles and their class names and dynamically insert the classnames with prop based values. Another approach could be to extract the component specific elements into separate component wrappers around the common date filter... whatever is most readable and reusable as a practice for future refactoring.

necoline:
Okay, so our dropdown component that we are using for dates and selects is a bit of a beast. (see https://github.com/IFRCGo/go-frontend/blob/5d74fd2c0ebe119cf18b00c4b203da99babab2d2/app/assets/scripts/components/dropdown.js#L32 
After spending some time trying to figure out how it worked so I could make modifications to apply it to the map selects, I backed out.

To deal with the styling, I took the first approach mentioned above of applying styles specific to the dates feature type. 

The dropdown has strange behavior that triggers a re-render when filters are submitted, this wasn't working on the map date filter, rather it was stuck in an open state. Neither of theses scenarios is ideal. 

I implemented a brute force method of closing the date filter after submission.

I think this fix is adequate for now. I'm not happy with the variations of selects that we are using across the site and would be more than happy to find a replacement to our dropdown component. While it's widely used across the site, I think we think about making plans to update this to something a bit more wieldy.

necoline:
Adds date filter to maps on home and region pages
## Description
I applied the date filter from the table to the map and made the dropdown styles dynamic to remain consistent with neighboring selects

I did this by extracting the date filter from the `display-table` into a common component accessible to all components making use of it. 

## Related Issue
#828 

## Motivation and Context
The data populating the map and table is the same. While some of the logic needed to implement the tables and map is feature specific, there is data that should be displayed through both mediums. Allowing the users to select date ranges for the map display was part of this process. 
As I made this available, I also tried to make steps to create common components between the map and table so we can work towards offering more common filtering between the components

## How Has This Been Tested?
This has been manually tested in development


Home:
image

Region:
image

## Types of changes
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

batpad:
@necoline the implementation looks good to me and the date filter works well.

>  I think it would worthwhile to discuss a replacement for this dropdown and unification of selects across the site.

It would be great to create a ticket for this - this would be a good conversation to have with @frozenhelium as well and do a general assessment of the state of dropdowns across the site and how we want to standardize.

@necoline small feedback from a user perspective, not sure of the best way to fix:

 - The `hover` styling for the Date drop-down seems a bit inconsistent with the filters next to it (turns grey)
 - The `Date` heading jumps a little bit when clicked on

^ These are probably the result of some default styling for `<a>` elements? These are not a huge deal.

The other thing is I wonder if we need some sort of indication that the Date filter is currently active. For the filters next to, for example, you see the currently searched for item. With the Date filter, currently, there's no obvious way to know if the filter is active or not without clicking on it. This maybe a UX question for later, though.

Overall, this looks good, and structurally sets us up well to make the filters across the table and map consistent as well as make this whole map + table + filters behaviour consistent across the home and region pages.

batpad:
@necoline I left a note about the logic of the date comparator. If we can get that resolved, I don't think any of the other issues should block merging this.

PR #841, opened 2019-10-15
Fix/per form submit

batpad:
cc @GregoryHorvath am creating a Pull Request for this.

Just testing it locally to see if all the scenarios work.

### DIFF
--- a/app/assets/scripts/components/per-forms/per-form-component.js
+++ b/app/assets/scripts/components/per-forms/per-form-component.js
@@ -2747,12 +2747,22 @@ const newFormElementKey = () => 'form' + Math.floor(Math.random() * 99999) + Dat
 
 export const PerFormComponent = (props) => {
   let button = null;
-  button = (<React.Fragment>
-    <div className='text-center'>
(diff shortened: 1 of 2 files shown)


batpad:
@GregoryHorvath -

So, now I can Save a Draft, then Edit the Draft, and it will allow to Save the Form, and Saving the Form works :+1: .

However, the Draft I have saved still exists - not sure if the Draft should be removed once the user has saved the form.

Also, the newly saved form does not show up in the `Active PER Forms` until I reload the page.

I am fine with merging this fix and looking at these two issues separately, just wanted to flag that those 2 issues still exist for me.

GregoryHorvath:
@batpad 

> Also, the newly saved form does not show up in the Active PER Forms until I reload the page.

Strange, this never happened for me. It simply redirects me back to the Account page with the PER FORMS tab and I have the updated list immediately.

> However, the Draft I have saved still exists - not sure if the Draft should be removed once the user has saved the form.

I've made a TODO comment about this in `per-form.js` if we should have a delete function for the drafts, or maybe getting them deleted after a submit automatically. I don't know who should be responsible for a decision like this.

PR #843, opened 2019-10-17
[WIP] Delete PER draft

batpad:
Making this a Pull Request so it is easier to organize comments @GregoryHorvath 

WARNING: This PR should not be merged until the backend branch to implement deletion is merged and deployed: https://github.com/IFRCGo/go-api/pull/534

@GregoryHorvath I pushed a fix for the issue of the page not refreshing the draft list after you delete.

All this PER code could probably be structured better and this is not ideally how I'd do it, but this works - let me know if you want to talk a bit about that code and I can explain exactly what it is doing.

### DIFF
--- a/app/assets/scripts/actions/index.js
+++ b/app/assets/scripts/actions/index.js
@@ -513,6 +513,11 @@ export function deletePerWorkplanApi (payload) {
   return postJSON('api/v2/del_perworkplan/', DELETE_PER_WORKPLAN_API, payload, withToken());
 }
 
+export const DELETE_PER_DRAFT = 'DELETE_PER_DRAFT';
+export function deletePerDraft (payload) {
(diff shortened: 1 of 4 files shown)


batpad:
@GregoryHorvath do you know if the backend changes for this have been deployed to staging? Then I think we can merge this.

GregoryHorvath:
> @GregoryHorvath do you know if the backend changes for this have been deployed to staging? Then I think we can merge this.

Yes, I believe I notified you by e-mail or comment or something about this (I forgot where exactly). @szabozoltan69 deployed it to staging for sure and maybe even for production as well I don't know, so this should be good for deployment.

szabozoltan69:
The back-end changes are kept synchronized in both staging and production recently, yes.

Issue #849, opened 2019-10-17
Graphs show exact data as home-page but filtered by region.


necoline:
## Issue

The chart on the region does not show correct data points.

## Current Behavior
The chart on the region page show operations over the last 10 years
![image](https://user-images.githubusercontent.com/20410256/67032174-0c871d00-f0e1-11e9-9eb0-9cf60a20f4ed.png)

## Expected Behavior

The charts should show appeals and DREFs over the last year and over the past decade
![image](https://user-images.githubusercontent.com/20410256/67032204-1b6dcf80-f0e1-11e9-87bc-56f100693ddd.png)

Related Issues: #851

## Description
Key figures are referred to as [homecharts](https://github.com/IFRCGo/go-frontend/blob/develop/app/assets/scripts/components/homecharts.js) in the code base

## Tasks
- [x] change the homecharts to a common component
- [x] create filter for region
- [x] add charts to region page

LukeCaley:
User test accepted (@mmusori and @LukeCaley).  Proceed to deploy.

PR #864, opened 2019-11-11
Fix/849 graphs

necoline:
## Description
- Adds region parameter to request for aggregate chart data
- Moves all logic related to the chart to the chart component
- Renames `HomeChart` to `TimelineCharts`
- Applies `TimelineCharts` to region page
- Deletes old region chart

## Related Issue
#849 

## Motivation and Context
The region page redesign requires different chart data to the one pre-existing set of charts

## How Has This Been Tested?
- Tested in the browser

Home page (pre-existing behavior)
image

Region page (Africa):
image

Region page (Europe):
image

## Types of changes
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)


batpad:
@necoline looks good!

Can we just move the graphs on region page to just below the Map + Operations Table (instead of below the country list) - this will keep it consistent with the home-page.

Issue #833, opened 2019-10-03
Field Report Early Warning option


batpad:
On the Field Report form, allow user to choose `Early Warning` as the field report type.

In this case, various field options and help texts need to change to be consistent with the Early Warning type. These are outlined in the Invision design.

So, we need to add the Early Warning option, and then go through the Invision doc, identify all the things that need to change based on the Field Report type selected by the user, and then implement these changes.

batpad:
@mmusori

batpad:
This is on hold until we get clarification on all changes and final functionality specs from @mmusori .

batpad:
Thank you @mmusori for the detailed email with changes from the wireframes. Pasting your email below (with a little added formatting):

### Context page

 - Status: should read “Early warning message for the risk of an upcoming disaster”
 - Start date: would propose it be worded more clearly as “Predicted Date of Impact”


### Situation page

 - For the sources, change to “Red Cross/Red Crescent”
 - Situational Overview: change to “Risk Analysis” and for the explanation to be changed to “Brief overview of the potential disaster and projected impacts”
 - Optional box where to place a weblink to the source is referred to, in order to be able to verify the figures given (and avoid a random figure is provided).


### Early action page

 - For “Assisted by Government and RCRC Movement”, add  “Number of people …” at the beginning to read “Number of people assisted by Government” and “Number of people assisted by RCRC Movement”
 - For early actions taken by NS, would suggest to:
    - Add “Activation/mobilization of volunteers”
    - Add “Public awareness raising and sensitization”
    - Add “WASH”
    - Add “Food security and livelihoods”
    - Remove “Community engagement activities” – broad, unclear
    - Remove “Community early warning” – 
    
 - For early actions taken by IFRC, would suggest:
    - Add,  “Support NS submission to FbA by the DREF or Imminent DREF” 
    - For early actions taken by RCRC Movement, see comments for NS. Change this to Early Action taken by other RCRC Movement
    - Add “Activation/mobilization of volunteers”
    - Add “Public awareness raising and sensitization”
    - Add “WASH”
    - Add “Food security and livelihoods”
    - Remove “Community engagement activities” – broad, unclear
    - Remove “Community early warning” – 

 - For Early Action taken by others, also ask to specify who the others are.

 
### Response page

 - Change “Trigger-based Early Action” to “FbA by the DREF” .
 - Add another option called “Forecast-based Action
 - Change “DREF Requested” to “DREF”
[...]

PR #876, opened 2019-11-26
Implement changes for Early Warning Field Report

batpad:
Refs #833 

Implements all conditionalities and changes required for the Early Warning Field Report.

Additionally, re-organizes how Actions are rendered - now rendering from data supplied by the back-end instead of hard-coded values on the frontend.

cc @necoline


batpad:
@necoline I've implemented all the changes except for the changes with the Action checkboxes, which is a bit of a larger restructuring.

I'll start working on that now, but if you could give this a look and let me know any feedback on approach, etc., would be great.

batpad:
@necoline this should be fully ready for testing and review.

necoline:
I think you have done a great job with this. It's great to see this code get cleaner with these additions. 👏 

There were a few things that I noticed:
- Step 2 is set up for only number inputs so the validation misses the text input for `affectedPopCentres` which should be a string
- this is a preexisting bug, but the check-box `Damage/Needs assessment`throws this error when checked: `TypeError: Cannot set property 'checked' of undefined` )  ... I don't know if this is something to add to this PR, but I wanted to make note of this so we can return to it later
- Upon completion, I cannot go back and edit the form: 
image

It looks like there is an error with the value being passed int `sourceEstimationMapping` in the utils

Just minor things aside from the edit button. Overall, really well done!

batpad:
Things TODO to complete this PR:

 - [x] Fix stepper titles according to form status
 - [ ] Fix validation for `Affected Pop Centres` field (needs to be a string)
 - [ ] Fix loading of hard-coded actions in `getInitialState` to use actions sent from the API
 - [ ] Fix `convertFieldReportToState` to correctly load field report for editing with new fields, etc.

There are two additional items todo to complete the Field Report Early Warning:

 - Add Yes / No field for `National Society Requires International Assistance` (similar to `Government Requires International Assistance` to appear right below) - this field appears on both EVT and EW forms.
 - Add a field for `other_sources` as a text field in Step 2, for the user to specify text for sources where they have mentioned `Other` as the source above.

I will create a separate ticket for these two additional fields ^ and do these in a separate PR.

batpad:
> Fix validation for Affected Pop Centres field (needs to be a string)

@necoline this is already defined as a string in the validation schema definitions. The validator does pass an all-numeric input as a "string". I do think that changing that is a bit more work and is not how string validation works in other places, so I'm inclined to leave this as is.

I think I've fixed all the other issues with loading the form, and the problem with some checkboxes. Loading of the Field Report depends on this backend PR being merged: https://github.com/IFRCGo/go-api/pull/555

batpad:
I believe this is ready to be tested. I have created a separate ticket for the additional two fields to be added: https://github.com/IFRCGo/go-frontend/issues/880

I'm going to squash commits here so we have a single commit for this.

PR #877, opened 2019-11-26
Implement 3w feedbacks

frozenhelium:


app/assets/scripts/views/ThreeW/project-form.js
```diff
@@ -509,6 +509,11 @@ class ProjectForm extends React.PureComponent {
         </InputSection>
 
         <footer>
+          <button
```
batpad: This seems like a reasonable hack :-) - can we add a comment in the code as to why this hidden button is here? Maybe difficult to figure out without context later.

frozenhelium: Sure, will do

PR #878, opened 2019-11-26
[WIP][Don't Merge] Feature/breaking out steps in EW /field report

necoline:
this is generally the pattern I was looking to pursue. There are some unresolved issues with passing through field change methods as props. Seeing looking at your work with the conditional fields, I think the splitting out of steps into components could work nicely with your pr


batpad:
@necoline am going to close this PR, but keep the branch around. This will be a great reference when we get down to tackling refactoring the field report form code.

Issue #847, opened 2019-10-17
Add Key Figures


necoline:
## Issue

Key figures should be present on region and country pages.
Only key figures relevant to the region or country should be present.
Styles for Key figures should be consistent across the site.

Related Issues: #851

## Description
Key figures are referred to as [homestats](https://github.com/IFRCGo/go-frontend/blob/develop/app/assets/scripts/components/homestats-component.js) in the code base

## Remaining Questions
- What do to with the country statistics that are not part of the key figures?
For example, the current design shows the `country statistics` and `national societies` along side key figures. With the implementation of the new design for key figures, we will need to place these statistics somewhere else from where they currently are.

![image](https://user-images.githubusercontent.com/20410256/67028265-00975d00-f0d9-11e9-95aa-40784fad5f50.png)

## Tasks
- [x] change the homestats to a common component
- [x] consolidate logic for filtering for key figures
- [x] extract tooltip to be separate component
- [x] add key figures to country page
- [x] add key figures to region page
![image](https://user-images.githubusercontent.com/20410256/67027652-e90ba480-f0d7-11e9-8e1d-fbbb0272e478.png)

LukeCaley:
@necoline The country statistics and National Society data appears instead in the country overview tab as per wireframes:  https://projects.invisionapp.com/share/K8R901UBXN4#/screens/356658486

![country-overview](https://user-images.githubusercontent.com/7791954/69621636-c312e180-103f-11ea-9247-06b821ecfee6.PNG)

LukeCaley:
Key figures need to replicate the content, text (i.e. not affected pop but targeted pop & funding requirements) and design as the GO home page.  These cards need to be filtered according to country and disappear when there's no data (as per expected behaviour across the site).

necoline:
Thank you for addressing the question of where to place the country key & national society indicators. I can add those back in at the specified location. 
Regarding the question of which key figures are shown on the region and country page, it was my understanding that the figures in the pre-existing view (affected people, requested amount, funding) constituted as 'key figures'

For clarifications sake, `key figures` on the home page are referenced as `appealsList` on the backend. Currently the region and country page are pulling from `appealsStats`.

`appealsList` is apparently set up to filter by country. It is less apparent on how one should filter by region. This will require further investigation. 

I can add the country key and national society indicators into the country overview tab and look into what it would take to change out the stats in the key figures for the region and country pages

batpad:
[...]

PR #882, opened 2019-12-02
Fix/847 update key figures

necoline:
#847 
- Found region filter for appeals list
- Created a separate action for appeals list stats because the appeals list table would access the same action and reset the region params causing the key figures to display the total appeals stats. I think there is probably a better way for handling this. I would be interested to hear further input. My experience with redux is rather limited so I went the verbose route to handle this issue 
- Cleaned up the common component since there was less variant logic to be handled. 
- Added country key statistics into a table
country page: 
image

region page:
image

image

home page:
image

image

country overview:
image


batpad:
@necoline I think you may have missed committing a file? I'm getting `Error: Cannot find module '../components/common/table-basic'`

PR #887, opened 2019-12-03
filter highlighted operations by is_featured_region flag

batpad:
This PR implements the filter to show Highlighted Operations on the Region page that are marked as `if_featured_region` on the backend.

@necoline I wasn't sure what the behaviour should be on the Country page. I have commented it out in the code for now, and would ideally just drop it from the country page, but we should discuss this.

I find it a bit awkward that I'm re-using the `GET_FEATURED_EMERGENCIES` action type in the action: https://github.com/IFRCGo/go-frontend/compare/feature/filter-region-ops?expand=1#diff-5c355adac55012fbbb2981b6d27b1742R166 - I think it's probably fine, but @necoline let me know if you think that should be done differently.

This seems to work and I tested locally with the backend. I am a bit confused by the `getFeaturedEmergenciesDeployments` action and the query it makes - it does not seem to apply any filters on the API, so have left it as is, but is possibly something to ticket out / improve later.

@necoline would be great to have a review.

TODO before merging:

 - [x] Decide what to do with Highlighted Ops on country page
 - [x] Remove commented out bits of code

### DIFF
--- a/app/assets/scripts/components/highlighted-operations/index.js
+++ b/app/assets/scripts/components/highlighted-operations/index.js
@@ -4,7 +4,6 @@ import { PropTypes as T } from 'prop-types';
 import { Link } from 'react-router-dom';
 import { environment } from '../../config';
 import { getFeaturedEmergencies, getFeaturedEmergenciesForRegion, getFeaturedEmergenciesDeployments, getDeploymentERU } from '../../actions';
-import { countriesByRegion } from '../../utils/region-constants';
 import BlockLoading from '../block-loading';
 import Fold from '../fold';
 import OperationCard from './operation-card';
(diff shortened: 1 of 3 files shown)


app/assets/scripts/components/highlighted-operations/index.js
```diff
@@ -75,12 +79,12 @@ class HighlightedOperations extends React.Component {
...
-    }
-    if (this.props.opsType === 'country') {
-      operations = operations.filter(op => op.countries[0].id === this.props.opsId);
-    }
+    // if (this.props.opsType === 'region') {
+    //   operations = operations.filter(op => countriesByRegion[this.props.opsId].includes(op.countries[0].id.toString()));
+    // }
+    // if (this.props.opsType === 'country') {
+    //   operations = operations.filter(op => op.countries[0].id === this.props.opsId);
+    // }
```
necoline: If the initial request in mount falls back to `_getFeaturedEmergencies` will we still need to access this logic?

batpad: It should never "fallback" - if the type is `region` it would query for emergencies filtered by region, and if that request fails for some reason, it will set the error state to true for emergencies which should then be handled the same way as errors for `_getFeaturedEmergencies`. Does that make sense?

batpad: I left these lines commented out in case we want to use this logic to still show "Highlighted Operations" on the country page, but those would have to be operations highlighted for the region - I don't think we want to add another "Highlighted for Country" flag in the backend. But thinking about this, I would really vote strongly to remove from the country page.

necoline: ah yes, sorry I missed the fact that we are are using these actions differently

PR #890, opened 2019-12-03
Fix type error in country page and key-figures-header

frozenhelium:

### DIFF
--- a/app/assets/scripts/views/countries.js
+++ b/app/assets/scripts/views/countries.js
@@ -462,7 +462,7 @@ class AdminArea extends SFPComponent {
       (typeof this.props.getPerMission !== 'undefined' && this.props.getPerMission.fetched && this.props.getPerMission.data.count > 0);
   }
 
-  renderCountryProfile () {
+  getCountryProfileData = () => {
     const { fetched, data } = this.props.fdrs;
     if (!fetched) {
       return null;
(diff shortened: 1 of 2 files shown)


app/assets/scripts/views/countries.js
```diff
@@ -664,20 +691,7 @@ class AdminArea extends SFPComponent {
...
-                      <BasicTable tableContents={this.renderCountryProfile().nationalSociety} tableTitle='National Society' />
-                    </div>
-                    <div className='table__basic-footer'>
-                      <p>
-                        <a href='http://data.ifrc.org/fdrs/' target='_blank'>
-                      Source: {this.renderCountryProfile().source.title}
-                        </a>
-                      </p>
-                      <p className='table__basic-footer-line'>
-                       | Reporting year(s): {Object.keys(this.renderCountryProfile().source.reportingYears).sort().join(', ') || 'N/A'}
```
frozenhelium: @necoline 
A few suggestions here:
* Should not call `this.renderCountryProfile()` multiple times for the same object but different property (It'll recompute the values every time), instead should have stored the result to a variable and accessed the properties through that variable.
* Should not directly access the property if it can be null (you've returned `null` if the `renderCountryProfile()`) 
* Should check for undefined or provide default value before calling its function (e.g. in `source.reportingYears.sort()`, `reportingYears` could be undefined which will cause an error. Instead we can check for undefined or use default value as `(source.reportingYears || []).sort()`

necoline: These are very helpful fixes. I really appreciate the review 👍

Issue #839, opened 2019-10-11
Consolidating map and table components and filters on Home and Region page


necoline:
## Issue

There is a fair amount of duplication across the the table and map features on the home and region page. 

## Expected behavior

The tables should share as much logic as possible and should make clear distinctions in data shared between the map and table to set this up for potentially creating a single filter system for both the table and map.

The map/table on region should be updated to include the same features present on the home page.
This means:
- [x] add presentation mode
- [x] remove bar chart
- [x] add legend
- [x] add filters with `in <Region_Name>`
- [x] add legend
- [x] update layout to match home_map
- [x] maintain automatic region filtering

batpad:
@necoline - just a clarification here:

We do want the map + filters + table on the region page to behave **exactly** like they currently are on the home-page - the behaviour and design, including the data displayed on the map, should be exactly the same. Except, of course, on the region page, all the data is filtered by the region one is currently looking at.

Ideally, this would also include the Presentation Mode functionality. So, this would mean, on the region page:

 - Make map on region page consistent with home-page (don't show bar graph on map, show legend on bottom-right)
 - Make layout consistent - i.e. View All Operations on top, then map and table as a single component. We may want `View All Operations` to say `View All Operations in <Region_Name>`, and it links, of course, to all operations, filtered by current region.
 - Implement filters on top of map exactly like home-page (except on the region page there is always an implicit additional region filter).

Essentially, literally everything should be the same on the region page as the home-page, except for the fact that all data is filtered by region on the region page.

Sorry for not clarifying this clearly earlier - let me know if there are any questions.

necoline:
Thanks for the clarification @batpad I'll revise my comment to include this

LukeCaley:
@necoline @batpad map and table filters are not linked on either home or regional page.  Complexity explained in https://github.com/IFRCGo/go-frontend/issues/828.  In order to implment this we would need to either a) resolve these complexities and find a solution which enabled the desired behaviour of the filtering options on the header of the map affecting both the map and table view simultaneously orb) we implement these filters independently and we provide a clear hint to the users that this is the case.  @batpad @necoline to clarify and propose solutions.

Secondary issues:
1) date format should be in sync, i.e. yyyy-mm-dd for both the filter and the table
2) hint for user that a filter has been applied in the filtering tab, i.e. slight change in colour?

necoline:
[...]

PR #892, opened 2019-12-04
Fix/839 update date table

necoline:
#839 
- Displays active date on date filter when date range is selected
- Does not change date format to yyyy-mm-dd since we are using html date input and that display change gets very tricky/convoluted very quickly. It looks like the general suggestion is to sub out the html calendar for a custom one with various date format display options. 
- Greys out inactive filter headers on table and changes color to black when active. I went this route because it looked clean to me. I think it could get busy if we changed the background color on active filters since you can have multiple active filters at once. I realize that using grey as the inactive color may convey that the filter is disabled, I am open to suggestions on how to improve this

table with active filters: type, name
image

date filter:
image

### DIFF
--- a/app/assets/scripts/components/common/filters/date-filter-header.js
+++ b/app/assets/scripts/components/common/filters/date-filter-header.js
@@ -11,6 +11,7 @@ export default class DateFilterHeader extends React.PureComponent {
     this.state = {
       startDate: null,
       endDate: null,
+      setDate: null,
       isClosingDropdown: false
     };
     this.resetDateStatus = this.resetDateStatus.bind(this);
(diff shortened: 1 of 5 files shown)


batpad:
@necoline this looks good.

The only problem I'm having with this visually is that the columns that are non-filterable are bold / highlighted the same as columns with an active filter. Can we make the non-filterable columns also grey, so they don't stand out and make it look like they are being filtered on (when they are non-filterable?)

Showing the active date on the date filter looks A+

necoline:
Thanks! Yeah, I wasn't sure about how to best handle the case of non-filtering headers from a visual perspective. I think your suggestion is the way to go

necoline:
I changed the color so that the default is black and the active filters are red:
image

Issue #844, opened 2019-10-17
PER: map to show PER phase instead of PER forms


teklal:
The map on the PER page (on staging as production site doesn't have PER data yet) currently shows information on the PER forms that have been completed by a National Society. The request from the PER team has been to have the map instead display the PER phase that the NS has last completed. 

![image](https://user-images.githubusercontent.com/52411237/66998588-cd52cf00-f0d4-11e9-9901-184922c578e5.png)

The colors of the dots on the map should correspond to the colors of the respective PER phase as shown in the image on the PER page (snip below):

![image](https://user-images.githubusercontent.com/52411237/66998859-43573600-f0d5-11e9-8c05-9062c6953713.png)

LukeCaley:
@necoline is this possible within the scope of the PER refactoring for GO 4.1?

necoline:
@LukeCaley I don't see this as being a very heavy lift, I think it would be nice to have this included to showcase improvements that are visually apparent to the user. I will carve out a few hours for additional enhancements upon completion of the PER update to add this in.

teklal:
@necoline did you have a chance to work on this yet?

necoline:
The entirety of my time last quarter was spent working on updating the per forms. I spent some time this week on this issue, the ticket above describes in detail what I did

teklal:
Thanks @necoline , I've taken a look at the current state of this on staging and it looks great! The one thing that I know has to be adjusted is the phrasing ("Last submitted phase" likely to be changed to "Current PER phase") on the legend, I am waiting for the confirmation from the PER team for their desired wording and will let you know as soon as I have that.

necoline:
Sounds great, that will be a quick adjustment to make when you are ready

teklal:
So the feedback from the PER team is to use "Current PER phase"

batpad:
Thanks @teklal - I have made this change in https://github.com/IFRCGo/go-frontend/pull/959 and merged to `develop`. This change should be on staging in a few minutes.

batpad:
This text change is done. Please re-open if anything needs to change.

teklal:
Hi @batpad , thanks for making the text changes. 

There is one thing that I've noticed with the map and that needs to be addressed:
The dot over Argentina is currently black (which is not a color in the legend and does not correspond to any of the phases) and I haven't figured out a reasoning for that color. As it doesn't have a phase specified, it should be gray i.e. "incomplete"

[...]

PR #941, opened 2020-01-24
changes layer to display phase

necoline:
## Description
Changes the map layers to show the phases rather than the section of the form completed. I added a 'incomplete' option which represents any form that has not completed any phase. I don't know if this is a realistic scenario. I know that some of our forms created for testing fall in that category but that might be due to the fact that these forms were not completed as intended. 

I think we should definitely have a fallback color in the case that users do not complete a phase, but I'm not sure if it should be included in the legend

## Related Issue
#844

## Motivation and Context

> The map on the PER page currently shows information on the PER forms that have been completed by a National Society. The request from the PER team has been to have the map instead display the PER phase that the NS has last completed.

## How Has This Been Tested?
 _Please describe in detail how you tested your changes._
_Include details of your testing environment, and the tests you ran to_
_see how your change affects other areas of the code, etc._

image

image

## Types of changes
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)

### DIFF
--- a/app/assets/scripts/components/map/per-map.js
+++ b/app/assets/scripts/components/map/per-map.js
@@ -64,9 +64,14 @@ class PerMap extends React.Component {
   }
 
   setMarkerLayers (data) {
+    const geoJSON = data.data.geoJSON;
+    geoJSON.features.map(feature => {
+      feature.properties.phaseCode = feature.properties.phase.phase.toString();
+      return feature;
+    });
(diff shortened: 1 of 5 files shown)


batpad:
@nanometrenat based on conversation, I'm merging this in. Please let us know any feedback.

PR #957, opened 2020-01-31
hotfix for #954 - preparedness tab does not render in production

### LINKED ISSUE #954: [PROD] PER:  Preparedness tab does not render in Production
## Issue
https://go.ifrc.org/preparedness does not function.
The issue arose after a user had submitted forms A3 and A3-2 for Barbados.

## Steps to reproduce
Log in and click on "Preparedness" at the top of the GO page.

## Expected behaviour
The Preparedness page should load.

## Related feature
PER #953 
PER #950 
PER #952 

As the issue arose after submitting forms A3 and A3-2 this problem may be/is likely to be related to the issues linked above

## Impact
Currently the problem affects all logged in users of GO navigating to the "Preparedness" tab.

## Severity
This should hopefully be fixed with the pending push to production of PER fixes.

## Is there a workaround?
There is no workaround.

batpad:
This adds a hotfix to filter out data entries where `region` is undefined, that was causing the preparedness page not to render on production.

Would be great to be able to merge this soon as it's causing an error in production - #954 .

@szabozoltan69 @GregoryHorvath do you mind giving this a quick look / test and let me know if this seems okay?

cc @nanometrenat - we'd like to merge just this change into the `master` branch and have it deployed on production, that fixes #954 .

### DIFF
--- a/app/assets/scripts/components/preparedness/national-societies-engaged-per.js
+++ b/app/assets/scripts/components/preparedness/national-societies-engaged-per.js
@@ -48,8 +48,8 @@ export default class NationalSocietiesEngagedPer extends PureComponent {
     const charts = [];
 
     // Filters out data where 'region' is undefined.
-    this.preparedData = this.preparedData.filter(d => d.region)
-    
+    this.preparedData = this.preparedData.filter(d => d.region);
+
     this.preparedData.forEach((region) => {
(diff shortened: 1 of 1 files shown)


batpad:
this was merged into develop and will go out as part of the next release. Closing this.

Issue #965, opened 2020-02-06
[PROD] Display change RR to Rapid Response in Deployments section and include in calculation


nanometrenat:
Logged on behalf of Sune

## Issue
Recently a new category of deployment was added - 'Rapid Response'.  Think it was via this change (not via a ticket):  https://github.com/IFRCGo/go-api/pull/590/commits/f5d16f41586f808b9a175ca2a9e8199950c15274#diff-cee64f30dcc9faf89588eda83273ec9a

1. On the front end this shows in the 'type' column as 'RR'.  This should show 'Rapid Response' NOT RR (It's fine for Rapid Response to be split across two lines).
2. In the Totals for deployed personnel, Rapid Response deployments are not included but should be.
3. 'Deployed FACTS' Label to be changed to 'Deployed Rapid Response' on Deployments page

## Steps to reproduce
1. Go to https://go.ifrc.org/deployments and look in Type column.
NB This data is also shown in other places e.g. https://go.ifrc.org/deployments/personnel/all and in the Emergencies page https://go.ifrc.org/emergencies/3972#details - the display should be updated in all pages where a deployments table is shown.
This is a display issue only, but it is important to use the correct terminology.
2. Where deployed personnel are totalled, these totals should include Rapid Response deployments in all the places where FACT is currently included.  e.g. the box for Coronavirus in Highlighted Operations on homepage and regional page currently shows 0 deployed surge personnel, even though Rapid Response personnel are currently deployed.
3. Go to https://go.ifrc.org/deployments - the top square box should include total of FACT+Rapid Response (not just fact and not just Rapid Response - as Rapid Response is the replacement terminology for FACT).  And label change needed as noted above.

nanometrenat:
Thanks, checked in Staging and Rapid Response is now shown in places 1 & 3 above, and the calculated totals include Rapid Response as well as FACT.   Good to go to Production :)

teklal:
Double-checked this, and as @nanometrenat said, Raid Response works in places 1&3 now as should.

One thing I noticed, where this is not yet included is the filtering by type in the Surge Alerts section. In the backend there is the option of choosing Rapid Response for the Alert Type but on the frontend it does not show as one of the options in the dropdown. 

![image](https://user-images.githubusercontent.com/52411237/76871186-0ac9ad00-686b-11ea-9fab-78cb1c18f4e9.png)

[...]

PR #967, opened 2020-02-10
Add FACT and Rapid Response values together

GregoryHorvath:
Issue: #965 

Changes:
- 'RR' to 'Rapid Response' in the Deployments table.
- Adds 'RR' and 'FACT' values together on the Deployments page and in the Highlighted Operations on the homepage.

Before merging:
- [ ] Check if I changed the right values for summing together 'FACT' and 'RR' numbers.
- [ ] Check if I missed any page where these values are used.

### DIFF
--- a/app/assets/scripts/components/connected/personnel-table.js
+++ b/app/assets/scripts/components/connected/personnel-table.js
@@ -149,13 +149,18 @@ class PersonnelTable extends SFPComponent {
         label: <SortHeader id='emer' title='Emergency' sort={this.state.table.sort} onClick={''} /> // for filtering options check .../api/v2/personnel/?limit=2 - the Filters button, Ordering
       }];
 
+      // Should add the other types if needed
+      const typeLongNames = {
+        'rr': 'Rapid Response'
+      };
+
       const rows = data.results.map(o => ({
         id: o.id,
         startDateInterval: DateTime.fromISO(o.start_date).toISODate(),
         endDate: DateTime.fromISO(o.end_date).toISODate(),
         name: o.name,
         role: get(o, 'role', nope),
-        type: o.type.toUpperCase(),
+        type: o.type === 'rr' ? typeLongNames[o.type] : o.type.toUpperCase(),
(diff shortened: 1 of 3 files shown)


app/assets/scripts/components/connected/personnel-table.js
```diff
@@ -149,13 +149,18 @@ class PersonnelTable extends SFPComponent {
         label: <SortHeader id='emer' title='Emergency' sort={this.state.table.sort} onClick={''} /> // for filtering options check .../api/v2/personnel/?limit=2 - the Filters button, Ordering
       }];
 
+      // Should add the other types if needed
+      const typeLongNames = {
```
batpad: Does this RR type come from a mapping on the backend or so? Can we add a quick comment pointing to the code in the backend if it does?

batpad:
This looks great to me - sorry for the late review @GregoryHorvath - have made a small comment - not sure if the `types` mapping there is something that comes from the backend - if so, would just be nice to drop a small comment in the code - if not, no worry.

batpad:
@nanometrenat can check if the business logic is as expected once this is merged (@GregoryHorvath feel free to go ahead and merge, I approved the PR). Looking at the code, the changes seem in line with what I think the expectation here is.

app/assets/scripts/components/connected/personnel-table.js
GregoryHorvath: It's here but I guess this is for the dropdown on the Django Admin only: https://github.com/IFRCGo/go-api/blob/e92b0ceadd70297a574fe4410d76eb7bf8614411/deployments/models.py#L98

batpad: Yup, if you can add a comment in the javascript code, something like:

```
// These types reference types defined in the backend models here: https://github.com/IFRCGo/go-api/blob/e92b0ceadd70297a574fe4410d76eb7bf8614411/deployments/models.py#L98-L106
```

Will just help if someone in the future needs to look-up where the `rr` value comes from. Just a code comment though :-) - the implementation looks fine to me.

Issue #973, opened 2020-02-20
[Staging] PER Forms being submitted with country_id as null


batpad:
## Issue

It seems like it is possible to submit a PER form such that the country id gets submitted as `null`. The steps to reproduce seem pretty reliable / am able to reproduce consistently.

## Steps to reproduce

 - Go to Account Page -> PER Forms.
 - Select a National Society from the dropdown
 - Select a Form to to create, eg. Area 1: Policy and Standards
 - Fill in form and submit

**Expected**: Form to be submitted with the correct country_id for the National Society chosen by the user.

**Actual**: The Form gets submitted to the backend with country_id as null, and the db does not validate presence of the country_id and saves the form, not associated to any country_id

Example JSON sent by the frontend when submitting a form:

```
{
	"code": "a1",
	"name": "Policy strategy form",
	"language": 1,
	"started_at": "2020-02-20 13:9:57.26+00",
	"submitted_at": "2020-02-20 13:9:57.26+00",
	"user_id": 3761,
	"ns": " - ",
	"data": [{
		"id": "a1",
		"op": 0,
		"nt": ""
	}, {
		"id": "c0q0",
		"op": 0,
		"nt": ""
	}, {
		"id": "c0q1",
		"op": 0,
		"nt": ""
	}, {
		"id": "c0q2",
		"op": 0,
		"nt": ""
	}, {
		"id": "c0q3",
		"op": 0,
		"nt": ""
	}, {
		"id": "c0q4",
		"op": 3,
		"nt": ""
	}, {
		"id": "c1q0",
		"op": 0,
		"nt": ""
	}, {
		"id": "c1q1",
		"op": 0,
		"nt": ""
	}, {
		"id": "c1q2",
		"op": 0,
		"nt": ""
	}, {
		"id": "c1q3",
		"op": 3,
		"nt": ""
	}, {
		"id": "c2q0",
		"op": 0,
		"nt": ""
	}, {
		"id": "c2q1",
		"op": 0,
		"nt": ""
	}, {
		"id": "c2q2",
		"op": 0,
		"nt": ""
	}, {
		"id": "c2q3",
		"op": 0,
		"nt": ""
	}, {
		"id": "c2q4",
		"op": 3,
		"nt": ""
	}, {
		"id": "c3q0",
		"op": 0,
		"nt": ""
	}, {
		"id": "c3q1",
		"op": 0,
		"nt": ""
	}, {
		"id": "c3q2",
		"op": 0,
[...]

PR #978, opened 2020-02-27
ref #973 - fixes cases where country id would be null on per form submission

batpad:
Fixes PER form to get country id correctly.


batpad:
Am going to go ahead and merge this to facilitate testing. @necoline would be great if you could give this a quick look though. Thanks.

PR #980, opened 2020-03-02
fixes #844 - black dot for phase 0 instead of grey

### LINKED ISSUE #844: PER: map to show PER phase instead of PER forms

batpad:
Fixes bug where countries that had their phase as 0 showed up as black, when they should be incomplete / grey.


batpad:
cc @necoline

PR #1029, opened 2020-03-19
[PROD] Handled if any of the Deployments RRs are undef.

GregoryHorvath:
Ref: https://github.com/IFRCGo/go-frontend/issues/965#issuecomment-601071308


batpad:
Good catch, @GregoryHorvath

PR #1031, opened 2020-03-19
Add project deletion feature

frozenhelium:


batpad:
@frozenhelium this looks really good to me. Can we:

 - Setup a "main" branch for the 3w / databank features (I think we can do one branch for both, at this point separating them out seems more work than it's worth)
 - Uncomment the 3w / databank code in that branch
 - Setup a surge.sh testing URL for the main 3w / databank branch
 - Merge all changes like this into the main 3w / databank branch, allow people to test on the surge.sh URL.
 - Only when we're ready, we merge to develop (if we only want to merge one of 3w / databank, we can then deal with commenting out one of them in the code.

@frozenhelium let me know if it helps to chat / run through any of the above. It will be really really nice to stay consistent and keep `develop` in a state that's always fit to deploy. Thanks!

frozenhelium:
@batpad Sounds great, do you have any naming convention pattern for such long running branch? Or should we just continue with the `feature/` convention?

Issue #1008, opened 2020-03-15
COVID-5 // Response docs box config - naming, pinned docs, ordering


LukeCaley:
**Problem**
Response document box names / categories do not work for COVID-19.  Renaming or adding boxes is not currently an option because this cascades throughout the application on every emergency page.  

Further, there is currently no option to pin key documents at the top of the box.  Finally, the order in which the response docs appear in each box is based on chronological order of upload rather than an order which helps to sort and find (e.g. alphabetical) or ensure key information is found easily (e.g. pinned docs or highlighting etc).

**Solutions**
The Response docs section to have the option of being configurable to specific emergencies, without cascading those changes throughout the site.

Additional back-end functionality introduced to allow users to pin documents, rename response doc categories if not fitting the default six provided and set the order of documents/links as they appear.

![COV-snip-response-docs](https://user-images.githubusercontent.com/7791954/76710642-7267e700-6709-11ea-8458-706c39694454.PNG)

xref: https://github.com/IFRCGo/go-frontend/issues/773 for the original requirement.  Also to note that this could help to resolve issues with COVID-19 reference docs, explained here: https://github.com/IFRCGo/go-frontend/issues/1007

batpad:
The Situation Report Types that are used as the headings for the Response Documents on the Emergency page can be added to in the database and users could upload documents to different Situation Report Types.

For individual Documents, we could add a field in the backend like `is_pinned` and handle showing of `is_pinned` items on the top. So, from a data entry and backend perspective, I think this should be relatively straightforward, both for users to add different title / groupings, as well as mark something as `pinned` within a certain grouping.

On the frontend, we currently always show 6 boxes, with the assumption that we will always have these 6 situation report types. The intention here was to also show these as empty if there was no documents for that type. Here, it seems like we have a few options:

 - Show box titles as per Situation Report Types added for documents in the backend. I.e. if an Emergency has 3 different Report Types, it will show 3 boxes - and this is based on whatever response documents and types are defined by users in the backend. This then becomes completely dynamic based on types and documents the user uploads, but we lose the ability of showing these standard boxes on other emergencies stating `No Documents Added`
 - The other option would be something like treating these 6 as default, and showing a `No Documents Added` if one of these headings does not have any documents. But, allowing users to add arbitrary Situation Report Types in the backend, and have those show up as additional headings / boxes on the frontend.

[...]

PR #1042, opened 2020-03-24
Response docs ordering

geohacker:
This PR should go along with https://github.com/IFRCGo/go-api/pull/639
* [x] Sort primary response docs with the specified order
* [x] Add non-primary response docs sorted by timestamp docs
* [x] Pin documents to the top of the list if `is_pinned`

cc @batpad


geohacker:
@GregoryHorvath @batpad could you merge this PR and then deploy to staging? This is required for the response docs to work well. Only part of it merge at the moment. This will also resolve the issue @teklal is facing https://github.com/IFRCGo/go-frontend/issues/1008#issuecomment-607748845

Issue #915, opened 2019-12-12
[PROD] Funding information incorrect on country page card


teklal:
The funding coverage card on the country pages is not showing the DREF fund coverage info.

![image](https://user-images.githubusercontent.com/52411237/70695711-965afd00-1cc2-11ea-85f3-a69657cefd87.png)

On the emergency page the cards do show the information correctly.

![image](https://user-images.githubusercontent.com/52411237/70696198-7ed04400-1cc3-11ea-8ccf-8e5a6c5bed9b.png)

Based on another country page's information it seems the problem indeed is in just the DREF funding coverage. As below, the funding requirements are a sum of both the DREF and the Emergency Appeal but the coverage takes into account only the EA.

![image](https://user-images.githubusercontent.com/52411237/70696974-023e6500-1cc5-11ea-9a41-0e8bd3282a9b.png)

Could we fix this so that it shows also the DREF funds info on the coverage card?

I noticed this on a country page, but I guess we may have the same problem/effect on the regional pages and potentially the home page cards as well?

nanometrenat:
Potentially related to changes made under https://github.com/IFRCGo/go-api/issues/536 ?

GregoryHorvath:
@nanometrenat So it won't add together the Funding for the highlights for DREF Appeals: https://github.com/IFRCGo/go-frontend/blob/866c5f490513465396daa79628af779a111fbe3a/app/assets/scripts/reducers/overall-stats.js#L52

@necoline (FYI) related commit: https://github.com/IFRCGo/go-frontend/commit/24ff1bc109fd8a0f6ae54a9675bf98f24560cd6f

necoline:
I went through past commits to see if somehow this data point was lost in past work on this feature.
It turns out that there was only ever one [endpoint](https://github.com/IFRCGo/go-frontend/blob/6217edd140a5c32cb06d905a0804f8e35097ed60/app/assets/scripts/actions/index.js#L151) provided for this feature ( #847) 

Since this data for this point on DREF Appeals is missing, we should look to the backend to see where this data is being supplied and find out why it is not available on the same endpoint with the rest of the data.

@GregoryHorvath 
   Do you have insight into why DREF Appeals are not coming through this endpoint? 
   Is there a reference point where the funding coverage for DREF Appeals is coming through where I could alternatively pull this data from?

GregoryHorvath:
@necoline This should be query firing (except the `time` part of it):
[...]

PR #1043, opened 2020-03-24
decouples the budget calculations from Appeals

necoline:
#915 
The funding calculations were applied specifically to Appeals and not to DREFs. I am unclear as to whether this was due a miscommunication during ticketing phase [here](https://github.com/IFRCGo/go-frontend/issues/847) or simply a code error.

This change moves the funding calculation to be applied to each object in the reduce rather than just the appeals.

**Note:**  The original logic for this component was copied from the region key figures. This change is applied to countries only and does not adjust the way the DREF and Appeals funding is calculated on the region page.

image

### DIFF
--- a/app/assets/scripts/reducers/overall-stats.js
+++ b/app/assets/scripts/reducers/overall-stats.js
@@ -54,17 +54,18 @@ function appealsList (state = appealsListInitialState, action) {
           if (endTime > now) {
             acc.activeDrefs++;
           }
+        }
 
         // Appeal.
-        } else if (object.atype === 1 || object.atype === 2) {
+        if (object.atype === 1 || object.atype === 2) {
           acc.totalAppeals++;
           if (endTime > now) {
             acc.activeAppeals++;
           }
-          const amountFunded = _toNumber(object.amount_funded);
-          acc.appealsBudget += amountRequested;
-          acc.appealsFunding += amountFunded;
         }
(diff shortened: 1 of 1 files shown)


GregoryHorvath:
@necoline Before I'd merge this, please confirm that this covers what Natalie mentioned in her comment here: https://github.com/IFRCGo/go-frontend/issues/915#issuecomment-603296107. That business logic seems a bit more complicated than what we currently have, but I've just ran through it quickly so I could be wrong.

batpad:
Could we put this up on a testing URL and ask @nanometrenat to confirm the calculation logic? @necoline

nanometrenat:
@mmusori is the man for the calcs.  Sorry I have no bandwidth for this kind of testing right now... Thanks loads for sorting!

necoline:
@batpad before we move onto testing, I want to make sure that we are clear on the desired behavior for this calculation 
> the business logic is not that Appeals and DREFS should be treated separately - in actual fact if there is an Appeal it can often require the DREF to be repaid. So if there is just a DREF then it needs to be the % of that that is calculated, but if there is an appeal as well then the calculations are different (depending on whether the funding amount is sufficient to repay the DREF)

there are still a few open questions on #915 that I would like to see addressed before we move forward with anything

batpad:
We may still have issues to tackle here, but this is being handled separately. Closing this as stale.

Issue #1013, opened 2020-03-15
Vanity URL


LukeCaley:
**Problem**
GO URLs are data-driven but quite difficult to remember.  It would be ideal if there was a way to create 'vanity' URLs to then be able to more easily redirect users verbally or without a hyperlink.

**Proposed solution**
Ability to give a name to the emergency page as a manual override to the sequential numbering, e.g. https://go.ifrc.org/emergencies/COVID

![COV-snip-URL](https://user-images.githubusercontent.com/7791954/76711142-9fb69400-670d-11ea-9f84-63d859a265db.PNG)

GregoryHorvath:
@LukeCaley Would we want to expand this to the whole site or just emergencies atm?

batpad:
What we should do:

 - Add a `slug` field to Emergency, where users can define a custom short name for the URL
 - Add the ability to query an emergency by `slug` in the API
 - In the frontend router, accept non-number routes for emergencies
 - When receiving a non-integer value in the route, fetch emergency by `slug`

We will probably need to do a bit of figuring out on the frontend to get this to work, but on the backend, this should be relatively straightforward.

LukeCaley:
Great, thanks @batpad 
This should be extended across the site but starting with ad-hoc emergencies, country and regional office pages.  So, for instance, I'm quite sure Regional Offices would be keen to have their region title instead of https://go.ifrc.org/regions/1#operations for example, so that they could share this more easily offline.

geohacker:
This is fixed through https://github.com/IFRCGo/go-api/pull/636. To test:

* Go to https://dsgocdnapi.azureedge.net/
* Edit an emergency, you'll see a `slug` field. Enter the vanity url slug 
![image](https://user-images.githubusercontent.com/371666/77154315-76a55300-6ac1-11ea-9021-630b3d39591f.png)

* Save
* Then go to https://dsgocdnapi.azureedge.net/emergencies/<slug-you-added>#details

cc @batpad

LukeCaley:
Assigned @teklal for testing and Q/A please

teklal:
Tested and works with limitations. @geohacker can you share (if known) what the limitations for choosing the slug are? 
Based on my testing I found at least these:
- the slug cannot start with a number 
- it is case-sensitive (would it be possible to make it not be case-sensitive, as I foresee this potentially causing a lot of confusion?)
- the slug cannot contain a # (didn't test other special characters. - and _ are fine, even at the start of the slug)

[...]

PR #1063, opened 2020-03-29
Fix/1013 slug requests

necoline:
#1013 
- creates child component for snippets
- moves snippet content from `admin-area-elements` to snippet component
- removes unused 'sitRepFilter'

image


geohacker:
This is deployed to https://ifrc-go-fix-1013-slug-requests.surge.sh/ for testing.

Issue #888, opened 2019-12-03
Handle different view on staging


batpad:
Right now, the site shows a discoloration on the staging site, and it makes the whole site look very different, making it harder to test some pieces.

Ideally, we should remove this huge discoloration, and just make an obvious change to the logo, or a space near the logo to clearly say `Staging` or `Testing` or `Alpha` or so.

TODO:

 - [x] Figure out how exactly the current setup is working / where the conditionals to swap styles out on staging are.
 - [ ] Remove the current conditional.
 - [ ] Agree on a much simpler indication of "Staging"
 - [ ] Implement the simpler indicator of "Staging"

cc @frozenhelium @necoline @LukeCaley

nanometrenat:
Not sure but this might perhaps be where the config points are for grey background, if that helps with step 1?:

https://github.com/IFRCGo/go-api/issues/518

https://github.com/IFRCGo/go-api/pull/538/commits/cf9dd9cb6a7528bca4d773d587a1e95ef4eb4756 https://github.com/IFRCGo/go-api/commit/cf05d70e0ff3146339046bbcd65f935d561113c6

batpad:
Thanks for digging that up @nanometrenat !

It maybe nice to streamline how we do this a bit, @thenav56 - is there a better pattern you'd suggest for doing this?

Next steps:

 - [ ] Design an element that we can toggle on / off with `display: none` or so to only show on the staging site. @nanometrenat are you able to help with a quick visual mockup (it can be very rough) of the desired behaviour here? i.e. how we want to indicate staging in the logo or so? (We can then adapt it to fit implementation)
 - [ ] Replace the logic in the build steps identified above to toggle the new element on / off instead of changing the background colour.
 - [ ] Look into improving how we do this conditional behaviour, it seems like the current way could be simplified and made a bit more obvious in configuration.

thenav56:
@batpad 

> Ideally, we should remove this huge discoloration, and just make an obvious change to the logo, or a space near the logo to clearly say Staging or Testing or Alpha or so.

I would also recommend this. Another will be using a bar fixed in top indicating server type.

We are using something like this in DEEP.

![image](https://user-images.githubusercontent.com/7059255/72072273-b1e31300-3315-11ea-911e-2c41353bd47b.png)
![image](https://user-images.githubusercontent.com/7059255/72072280-b3acd680-3315-11ea-820d-4b7aefab443c.png)

nanometrenat:
[...]

PR #1070, opened 2020-04-01
Removed current CSS, added banner for staging

GregoryHorvath:
Ref.: https://github.com/IFRCGo/go-frontend/issues/888

### DIFF
--- a/app/assets/styles/global/_pages.scss
+++ b/app/assets/styles/global/_pages.scss
@@ -187,7 +187,8 @@ main {
 }
 
 .sticky-banner {
-  height: 28px;
+  height: 32px;
+  font-size: 1rem;
   padding: 4px;
   position: fixed;
   width: 100%;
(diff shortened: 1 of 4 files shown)


app/assets/styles/global/_pages.scss
```diff
@@ -185,3 +185,16 @@ main {
     float: right;
   }
 }
+
+.staging-footer {
```
necoline, line 189: I think this is our only example of a sticky banner. What do you think of breaking out the styles into 
- sticky-banner
- staging-footer
So we can reuse this later for the new emergencies banner (or any other feature that might come along)

necoline:
This implementation looks good to me can you deploy to surge so we can test it out?

I don't know how well documented this is, but in case you need it: `yarn build-staging && yarn deploy-surge`

app/assets/styles/global/_pages.scss
GregoryHorvath: Check: 4053d2e

Issue #1009, opened 2020-03-15
COVID-6 // Add tab and rename


LukeCaley:
**Problem**
Users would like to add and rename tabs on Regional, Country and Emergency pages.  Currently there are differing numbers of pre-named and, to an extent, pre-populated tabs on each of these levels.  While this is true of the expanding range of data to be displayed for COVID-19, it is also true of regional and national pages.

**Proposed solution**
Users should be given the option to add a number of tabs (3 additional?) in order to provide them with flexibility to organise emergency, national and regional pages. 

**Alternatives considered**
The alternative might be to have an accordion style expansion based on different component areas.

Xref: https://github.com/IFRCGo/go-frontend/issues/855

*To note, there could be considerations around permissions for various tabs e.g. public vs only logged-in RCRC users able to see certain tabs*

![COV-snip-tabs](https://user-images.githubusercontent.com/7791954/76710804-b3143000-670a-11ea-827d-5d0c0387af93.PNG)

batpad:
@LukeCaley I'm fairly confused with the intention here. Users would add a tab, give it a title .. and then be able to insert arbitrary content? Or choose where current data from the tabs goes? Or just rename existing tabs to something else, but continue showing the data they currently show?

We could possibly think of having a feature where a user can add an arbitrary "Tab", give it a "Title" and some "Content", and display these additional tabs on the page if they exist. Does that sound like what we're after?

@LukeCaley it would be really great to get better context of the use cases here to be able to better suggest a path forward.

LukeCaley:
Thanks @batpad.  Indeed, this is "a feature where a user can add an arbitrary "Tab", give it a "Title" and some "Content", and display these additional tabs on the page if they exist".

One use case for an emergency might be https://go.ifrc.org/emergencies/3495#additional-info where this info might be found and organised more easily with tabs for `ops update` and `maps and graphics`.  

Another use case is for regional offices to be able to manage their information and specific requirements, for instance in the Americas where they have had to organise dashboards on KPIs, contact and office maps and a range of links into the same `additional-info` tab.

A further use case is for NS to add tabs dedicated to long-term emergency programmes, such as the Emergency Social Safety Net programme in Turkey, where the Turkish Red Crescent could dedicate a full tab to dashboards and other graphics.

nanometrenat:
[...]

PR #1071, opened 2020-04-01
Multiple tabs

geohacker:
Addresses #1009. Should be merged after https://github.com/IFRCGo/go-api/pull/650
@necoline  - let me know what you think, I think the way I'm managing the tabs in the state could probably be more elegant. 

One thing I'm struggling with which we should absolutely fix is this: https://github.com/IFRCGo/go-frontend/blob/feature/multiple-tabs/app/assets/scripts/views/emergency.js#L116 -- if we do this on componentDidMount, the tabs aren't populated yet so it always resolves to `#details`.

cc @batpad

### DIFF
--- a/app/assets/scripts/components/emergencies/snippets.js
+++ b/app/assets/scripts/components/emergencies/snippets.js
@@ -32,22 +32,31 @@ class Snippets extends Component {
 
 Snippets.propTypes = {
   _getEventSnippets: PropTypes.func,
-  eventId: PropTypes.string,
+  eventId: PropTypes.number,
   snippets: PropTypes.object
 };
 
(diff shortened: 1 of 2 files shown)


necoline:
Should we render all tab header and panels (aside from the default tab) in a child component?

geohacker:
@necoline I'm a bit hesitant to rework this into child components. I tried real quick and the way Tabs are defined in a TabsList etc seem a bit tricky to get right if we pull into child components. 

I just added a 0 set timeout after the tabs are registered and that seem to work well for users arriving at invalid hashes. I don't love this approach, but if you can think of better a way let's do that. Thanks!

cc @batpad

PR #1085, opened 2020-04-07
fix surge routing by using a 200.html

geohacker:
cc @batpad


batpad:
@geohacker this looks great. Can we do a deploy to surge from this branch and check that it works?

geohacker:
@batpad https://ifrc-go-develop.surge.sh/emergencies/3892#details

geohacker:
I deployed that before making a new branch, but it's the same change.

PR #1074, opened 2020-04-02
Feature/1004 epi field report

necoline:
#1004
The PR includes:
- disable early warning on epidemic disaster
- new fields for estimation source organizations
- epidemic specific situation fields
- date of data
- required start date
- validation for new fields 
- sections for action checkboxes on EPI (or any future form type that has multiple categories)
- larger text boxes

Known pre-existing issues:
- Minor form validation issues: fields accepting numbers do not check if number is an integer
- Does not reset the province field when the country is changed. 
- Does not allow ‘yes/no’ questions to be submitted as empty (will automatically choose ‘no’ if none is selected
- Does not add tool tip or other feature to explain abbreviations
- Does not automatically set title
	
Known EPI shortcomings:
- Does not include file upload (only text url) — recommended future work
- Does not update date of data on edit — will look into this

url to test: https://ifrc-go-feature-1004-epi-field-report.surge.sh/

PR #1103, opened 2020-04-14
Recommit lost changes

GregoryHorvath:


batpad:
This was due to some accidental merges in https://github.com/IFRCGo/go-frontend/pull/1074/

Issue #1097, opened 2020-04-12
[Staging] Epi Field Report figures don't show on Emergency page


nanometrenat:
## Issue
I created an Epi Field Report in staging however when I look at the associated emergency it still shows the old figure types (and not the epi figures)

i.e. Emergency Overview
AFFECTED--
INJURED--
DEAD--
MISSING--
DISPLACED--
ASSISTED--
LOCAL STAFF--
VOLUNTEERS--
EXPAT DELEGATES--
Source: Natalie Epi Field Report Sunday the Twelfth , 12/04/2020, 16:54

Whereas I was expecting
Cases (Ministry of Health):
56
Suspected Cases (Ministry of Health):
22
Probable Cases (Ministry of Health):
11
Confirmed Cases (Ministry of Health):
99

etc.

## URL
https://ifrc-go-feature-1004-epi-field-report.surge.sh/emergencies/3938#details

## Steps to reproduce
Submit an epi Field Report and check the numbers pull through

![image](https://user-images.githubusercontent.com/33625192/79071967-cb15aa00-7cde-11ea-9c9d-1e00a24098ad.png)

![image](https://user-images.githubusercontent.com/33625192/79071977-da94f300-7cde-11ea-8719-c9efab029bd9.png)

## Expected behaviour
I was expecting to see the Epi numbers

## Related feature
#1004 #1096

## Criticality/Urgency
Not a blocker for the release but should be assessed as to priority given current focus on epi.

nanometrenat:
@LukeCaley same with this re presumably being carried forward.  Not sure how you want to tag these to make sure they're not missed

GregoryHorvath:
@mmusori @nanometrenat Anything to change?
![image](https://user-images.githubusercontent.com/30079559/80228576-7fa0bb80-864f-11ea-9425-160aebdc62d4.png)

cc @batpad depends on: https://github.com/IFRCGo/go-api/commit/f1aea58e350f0c716e5e5a8195d43f908d9cffc9

mmusori:
After the above developments. The numeric details are not showing in the email notification and the emergency page.
![image](https://user-images.githubusercontent.com/7780809/80366964-85361580-888a-11ea-8007-3855007ac393.png)
The below is what comes out on the emergency page
![image](https://user-images.githubusercontent.com/7780809/80367123-cdedce80-888a-11ea-98d1-b32e07f747b4.png)

And this is the report generated 
[...]

PR #1151, opened 2020-04-27
Added EPI figures to Emergency page

GregoryHorvath:
Ref.: https://github.com/IFRCGo/go-frontend/issues/1097

Requires: https://github.com/IFRCGo/go-api/pull/680

### DIFF
--- a/app/assets/scripts/views/emergency.js
+++ b/app/assets/scripts/views/emergency.js
@@ -217,6 +217,31 @@ class Emergency extends React.Component {
             <li>Expat delegates<span className='content-highlight'>{n(get(report, 'num_expats_delegates'))}</span></li>
           </ul>
         </div>
+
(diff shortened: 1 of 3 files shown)


app/assets/scripts/views/emergency.js
```diff
@@ -217,6 +217,31 @@ class Emergency extends React.Component {
             <li>Expat delegates<span className='content-highlight'>{n(get(report, 'num_expats_delegates'))}</span></li>
           </ul>
         </div>
+
+        <h4 className='global-spacing-2-t clear'>Epidemic figures</h4>
```
batpad, line 221: @GregoryHorvath don't we need a conditional of some sort here so this only shows if the emergency type is Epidemic? I don't think we want this showing for all emergencies.

mmusori: That is the correct logic @batpad

PR #1200, opened 2020-05-06
Change PrivateRoute to Route for /reports/all

### LINKED ISSUE #1118: Field Reports page - auth required (public FRs should not require auth)
## Issue
As an unauthenticated user, when going to the Emergencies page and clicking through to view Field Reports (all regions or specific regions), it then prompts me to log in
https://go.ifrc.org/reports/all

I would expect that if I'm not logged in I would be able to see the list of field reports with visibility=public, and if I am logged in then I would see the list of field reports that I am authorised to view (Membership / IFRC)

Similarly, if I am looking at an Emergency page that has a linked Field Report then I am prompted to log in before I can view it.

Speaking on the GO call with @LukeCaley right now it sounds like this isn't a deliberate requirement to get users to log in in order to see PUBLIC field reports - so assuming Luke is happy then presumably we should be making the Public field reports Public.  Luke to confirm.

## Impact
Impact is that the general public can't see Field Reports even if visibility is set to Public.

kamicut:
Closes #1118

URL to test is here: https://ifrc-go-fix-1118-public-frs.surge.sh/reports/all

You can see that the response called returns objects with `visiblity` equal to 3, which means they are public reports.

### DIFF
--- a/app/assets/scripts/main.js
+++ b/app/assets/scripts/main.js
@@ -101,7 +101,7 @@ const Root = () => (
         <AnonymousRoute exact path='/recover-account/:username/:token' component={RecoverAccount}/>
         <AnonymousRoute exact path='/recover-username' component={RecoverUsername}/>
         <PrivateRoute exact path='/reports/new' component={FieldReportForm}/>
-        <PrivateRoute exact path='/reports/all' render={props => <Table {...props} type='report' />} />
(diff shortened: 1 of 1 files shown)


batpad:
Thanks for tracking that down and for the in-the-end simple fix, @kamicut :-)

This looks good to me. Let's wait for @nanometrenat and @karitotp to test and confirm this works as expected and we should be good to merge.

Issue #1112, opened 2020-04-15
[Staging] Epi field report: date of data field marked as mandatory but can proceed without filling it


teklal:
## Issue
The "Date of Data" field on the page with the figures on cases etc is marked as mandatory with an asterix but I am able to proceed without having filled this field.

As the fields with figures are not mandatory I'm not sure what the ideal solution is here - whether to remove the asterix or make filling at least one field with figures mandatory?

Thoughts @nanometrenat @LukeCaley ?

![image](https://user-images.githubusercontent.com/52411237/79341443-1670c800-7f2c-11ea-8b72-06a1b8859fe9.png)

## Steps to reproduce
Start filling an Epi FR and click "continue" on the page with the figures without giving a date.

## Related feature
epi field report and general field report issues

I don't see this as a major major problem but worth noting.

teklal:
Sorry @necoline, noticed this just after finishing the call!

nanometrenat:
There was a conversation about this on the wireframes - suggest taking a look at that, thanks @teklal

mmusori:
Proposed logic not yet implemented. If data is entered for cases or Dead then the date become mandatory. @GregoryHorvath We had a discussion about this.

batpad:
Just confirming this logic @kamicut - if any data is entered in the fields for Cases, Dead, etc. - then Date of Data must be filled in.

kamicut:
This is ready for QA in https://github.com/IFRCGo/go-frontend/pull/1281

LukeCaley:
@teklal and @karitotp to test

teklal:
Tested and seems to work now, thanks!

batpad:
This is now on staging.

karitotp:
This is working good!

LukeCaley:
@teklal can you check that this is working on PROD and close if so - thanks!

karitotp:
It is working as expected, it is mandatory when the numeric fields is fill.

teklal:
This works now, thanks!

PR #1236, opened 2020-05-15
Indicate that date of data is not required for EPI FRs

kamicut:
=> #1112 
surge: https://ifrc-go-fix-1112-date-of-data.surge.sh


kamicut:
There is a newer PR #1281 that addresses #1112

PR #1248, opened 2020-05-17
Add absolute import support for js files

frozenhelium:
Following is the resolution pattern
```
#actions: /src/root/actions,
#components: /src/root/components,
#config: /src/config,
#hooks: /src/root/hooks,
#lang: /src/lang,
#selectors: /src/root/selectors,
#utils: /src/root/utils,
#views: /src/root/views,
```
For example:
```js
import Header from '#components/header';
```
will work in any level without having to know the relative path for the `header.js`

Replaced all relative imports with absolute imports using following command
```sh
rg '(\.{1,2}/){1,}(components|actions|utils|config|hooks|root|selectors|views)' --files-with-matches | xargs sed -i 's/(\.{1,2}\/){1,}(components|actions|utils|config|hooks|root|selectors|views)/#\2/gm' -r
```


batpad:
This is really nice!

batpad:
@frozenhelium can we create a ticket to modify all current relative imports to use this pattern? we don't need to do them all right now, but if we have it ticketed, someone can easily take it on if they have a little bit of time. Would be nice to eventually be consistent to using this pattern everywhere.

frozenhelium:
@batpad Sure, I'll create the ticket once this is merged

Issue #1230, opened 2020-05-13
[PROD] PER: Enable dates beyond just 2018, 2019 and 2020 to be selected on Overview form


teklal:
## Issue
Currently the Overview form does not allow selecting a date from any other years except 2018, 2019 and 2020.

![image](https://user-images.githubusercontent.com/52411237/81840837-cfd1b600-9549-11ea-82f5-30b98b920c5b.png)

This issue applies for all the date fields in the Overview form.

It should be possible to choose any date from any year.

## Steps to reproduce
Go to Accounts page, PER tab and open an Overview form.

## Severity
This has been reported as a blocker for data entry by PER team.

nanometrenat:
I am intrigued that there are 3 dropdowns for each date - as in other places on the platform (e.g. Field Reports) there's a calendar date picker.  @teklal do you know if this approach was deliberate for PER (e.g. if it's easier for their data entry) or not?

teklal:
@nanometrenat that is an excellent question. As far as I know this was not a specific request/requirement. It is however possible (will have to check) that initially the desire was to have only dates as far back as 2018 as that's when PER as it is was "launched" and that may have influenced the choices made on how to implement the form and date picker - but those discussions were had prior to my time so not 100% sure on that.

teklal:
If changing the date fields to a calendar date picker is easy (doesn't change the way the dates are saved etc), then that would be a great way to solve this. Thanks @nanometrenat for the discussion on this.

karitotp:
Now the dropdown for year is from 2010 - 2023, but the icon of dropdowns are not shown as in the above image.

![Selection_003](https://user-images.githubusercontent.com/11504548/82324748-bee8d080-999f-11ea-9b5b-38447ed3d1ee.png)

nanometrenat:
@teklal can you please look at Karito's comment above and advise what actions are needed?

teklal:
Thanks @karitotp for the observation on the dropdowns! I propose we leave it as it is though as it's not a blocker and there haven't been any complaints from the users so far. It seems to be a design choice and applies to all the fields that have dropdowns on the overview form.

I'll leave this still open as there was a request to extend the range of years to go until 2030. @batpad would this be a quick/easy change or should I close this ticket and log that on a new one for future tackling?

batpad:
Have updated the list of years to extend to 2030. Available for testing here: https://ifrc-go-minor-per-date-range.surge.sh/

cc @karitotp @teklal

[...]

PR #1250, opened 2020-05-18
refs #1230 - add years from 2010 - 2023 in PER overview form years

batpad:
Allow years from 2010 - 2023 in PER overview form year dropdowns

Refs #1230


teklal:
Thanks @batpad, this is great! I just heard back from the PER team and could the range be extended to 2030? If not difficult to change at this point? 2023 is already great for now so if a complex change, it can wait for a later deploy/time.

PR #1254, opened 2020-05-18
Absolute imports/replace

GregoryHorvath:
Replacements done down until the `src/root/components/connected` folder.

### DIFF
--- a/src/root/components/connected/alerts-table.js
+++ b/src/root/components/connected/alerts-table.js
@@ -4,17 +4,18 @@ import { PropTypes as T } from 'prop-types';
 import { Link } from 'react-router-dom';
 import { DateTime } from 'luxon';
 
-import { environment } from '../../config';
-import { getSurgeAlerts } from '../../actions';
-import { get, dateOptions, datesAgo, isLoggedIn } from '../../utils/utils';
-import { nope, privateSurgeAlert, recentInterval } from '../../utils/format';
+import { environment } from '#root/config';
+import { getSurgeAlerts } from '#actions';
+import { get, dateOptions, datesAgo, isLoggedIn } from '#utils/utils';
+import { nope, privateSurgeAlert, recentInterval } from '#utils/format';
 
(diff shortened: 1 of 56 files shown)


src/root/components/connected/alerts-table.js
```diff
@@ -4,17 +4,18 @@ import { PropTypes as T } from 'prop-types';
...
-import DisplayTable, { FilterHeader } from '../display-table';
-import BlockLoading from '../block-loading';
-import Fold from '../fold';
-import Expandable from '../expandable';
+import { environment } from '#root/config';
+import { getSurgeAlerts } from '#actions';
+import { get, dateOptions, datesAgo, isLoggedIn } from '#utils/utils';
+import { nope, privateSurgeAlert, recentInterval } from '#utils/format';
+
+// FIXME: imports from the /components/ could be a 1 liner?
```
GregoryHorvath, line 12: @frozenhelium some of these could be 1 liners in most files?

frozenhelium:
@GregoryHorvath Wow, this is amazing! :+1:

src/root/components/connected/alerts-table.js
frozenhelium: How do you mean?

GregoryHorvath: Something like (could be wrong, didn't have the time yet to mess with this much):

```javascript
import ExportButton, DisplayTable, FilterHeader, ...etc from '#components';
```

`FilterHeader` might not work like this since it's not a default export, but somehow like this.

frozenhelium: That's true, we can add a `/components/index.js` file and re-export every components from there.  But I think, it's better to import components separately. Lets see if we can do that later.

GregoryHorvath: I don't know if that's actually necessary, like default exports should work simply like this as far as I know. I'm just not 100% sure. Have to test it later.

Issue #1232, opened 2020-05-14
Remove "Additional graphics" text from additional tabs


teklal:
As there is now the possibility to add more tabs (#1009) to emergency pages and these additional tabs can be used for sharing not only graphics but also guidance etc. I think we should remove the text "Additional graphics" from the top of these extra tabs as it does not anymore add any extra value or information as to the content of the tabs. The text currently appears for all the additional tabs.

![image](https://user-images.githubusercontent.com/52411237/81908888-1f57c680-95ca-11ea-9e3e-58b5d02e3ba5.png)

This is a more of a cosmetic issue and should not be prioritised for now if it's not a quick fix.

batpad:
@teklal this can be tested at https://ifrc-go-fix-1232-remove-additional-graphics.surge.sh/ - please let know if this works as intended from a business logic perspective, and we can ask @karitotp to go ahead and do a round of testing to see if there's any edge cases or bugs. Thank you! cc @kamicut

teklal:
Thanks, looks good to me. Would be good to have tested by Karito :)

karitotp:
ADDITIONAL GRAPHICS does not appear any more in the new tabs added when I add something in the snippet box, but when I left it creating only the tabs an not adding anything as content for that, the text "Additional graphics" is still shown. Is that the expected behavior?

teklal:
Thanks for that catch @karitotp. I don't think the text "Additional graphics" should be there when the tab is empty (as indeed there are no additional graphics shown in there in those cases :D).

GregoryHorvath:
@teklal @karitotp Is this also for the default "Additional Information" tab or just the other ones?
@batpad the PR: https://github.com/IFRCGo/go-frontend/pull/1279 removes the header for all of them. Wondering if we should keep that logic at all and remove the header code alltogether at this point?

karitotp:
@GregoryHorvath @teklal, "Additional graphics" text is appearing in the default "Additional Information" tab too.

GregoryHorvath:
> @GregoryHorvath @teklal, "Additional graphics" text is appearing in the default "Additional Information" tab too.

Sorry, my question was misleading. Do we want to remove from all of them or keep it for the "Additional Information" tab? This wasn't really clear for me.

LukeCaley:
[...]

PR #1279, opened 2020-05-21
Hide header if there is no data

GregoryHorvath:
Ref.: https://github.com/IFRCGo/go-frontend/issues/1232


kamicut:
Before:
image


After:
image

@batpad if this is the correct behavior for a tab with no data, then we can merge this.

PR #1281, opened 2020-05-22
Fix for mandatory date on step 2 of form

kamicut:
Link: #1112

This PR creates a dependency between the fields in the reported numbers and the date of data, such that you can't go past step 2 if you've filled the reported numbers field.

@batpad there is something weird about the schema in the number of reported dead, because the form has a field called `epiNumDead`, but the schema doesn't have this property, it has `numDead`. Thus I couldn't create a dependency on `epiNumDead`. Which one is the correct property? 

I am marking the PR as a draft until we figure it out. 

surge: https://ifrc-go-fix-1112-mandatory-date.surge.sh


batpad:
@kamicut yes, the field for reported dead on the epidemic field report is called `epiNumDead` - since it is not a required field, it is possible that it is missed from the schema, do feel free to add it (keeping in mind that this field does not show up in case this is a non-EPI field report).

kamicut:
@batpad I've added a link in the comment above

PR #1282, opened 2020-05-22
Add translation support to views/RegionalThreeW

samshara:


src/root/utils/lang.js
```diff
@@ -48,8 +48,7 @@ export function resolveToComponent(template, params) {
 
     return (
       <React.Fragment>
-        { part.replace(`${key}}`, '') }
-        { params[key] }
+        { part.replace(`${key}}`, params[key]) }
```
frozenhelium, line 51: This will convert components to string [object object]

Issue #1212, opened 2020-05-07
[PROD] Countries appear multiple times after saving in "Country-level notification" box


teklal:
## Issue
On the user account page (https://go.ifrc.org/account#notifications) there is the option to subscribe to notifications related to specific countries. When choosing a country and hitting save, these show up multiple times in the box.

Below a few examples (from different user accounts just to test that this is not specific to my account) I recorded to better demonstrate what I mean.

![countrynotifications2](https://user-images.githubusercontent.com/52411237/81290175-e1a5eb80-9067-11ea-9428-64690df6d324.gif)

![countrynotifications](https://user-images.githubusercontent.com/52411237/81289403-a7881a00-9066-11ea-9f10-ca5dfee014b2.gif)

## Steps to reproduce
- Navigate to the Account page and there to the Notifications tab (https://go.ifrc.org/account#notifications). 
- Choose one or more countries from the dropdown to the "Country-level notifications". 
- Hit save (bottom of the page).

## Expected behaviour
Country should appear only once in the box.

## Severity
I don't think (but haven't thoroughly tested) this affects the notification functionality so more of a cosmetic thing but would be nice to have fixed as it may appear confusing to users. Now that we are going to have a video explaining how to subscribe to notifications, there may be more users doing that.

GregoryHorvath:
@teklal @karitotp The fix is deployed to staging and should be testable.

karitotp:
Now this is working good to me!

teklal:
This is working as desired now, thanks!

PR #1285, opened 2020-05-25
Fix duplicated countries after subscr. saving

GregoryHorvath:
Ref.: https://github.com/IFRCGo/go-frontend/issues/1212


GregoryHorvath:
@batpad Here the issue was the same, that so many things relying on those static lists, like `countryList` in `field-report-constants.js`. Then it makes the `[{ value: '', label: ''}, ...]` arrays for the dropdowns from those but when it comes to those abstract functions which check if X is element of the list it "fails" because the `countryList` objects also have the `iso: ''` attribute in some cases.

PR #1286, opened 2020-05-25
Allow public FRs on emergency pages

kamicut:
To complete #1118 we need to be able to see public FRs on emergency pages in the "Field Reports" tab. I removed the "must login" logic, but I didn't know how to associate two FRs to one emergency to see what happens when there is a private FR as well as a public FR. Could you test @batpad @karitotp ?

https://ifrc-go-fix-1118-emergency-pages.surge.sh/emergencies/3833#details

### DIFF
--- a/src/root/views/emergency.js
+++ b/src/root/views/emergency.js
@@ -172,14 +172,6 @@ class Emergency extends React.Component {
     return false;
   }
 
-  renderMustLogin () {
-    return (
-      <React.Fragment>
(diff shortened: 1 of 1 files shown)


kamicut:
As I was testing, it seems that the API is returning a private field report along with an event for a non logged in user, here's an example https://ifrc-go-fix-1118-emergency-pages.surge.sh/emergencies/4013#details. Clicking on the field report will correctly block access to a non-logged in user

Public FR like the one above works, and doesn't block access for non-logged in user.

GregoryHorvath:
@kamicut @batpad @teklal Deployed the backend changes to staging so this should be testable. Now non-registered users should only see Public Field Reports, registered users should see Membership and Public Field Reports and the IFRC Admins / superusers should see all.

Issue #1141, opened 2020-04-23
[PROD] Field report: Possible to create field report without title


teklal:
## Issue
It is currently possible to submit a field report without a title if the field report is linked to an existing emergency page. The title is technically a required field and in all other scenarios besides when linked to an event, the system will prompt for a title before allowing a user to proceed filling the form.

![image](https://user-images.githubusercontent.com/52411237/80116296-16a14100-8586-11ea-94b5-435ba007673d.png)

On emergency page:
![image](https://user-images.githubusercontent.com/52411237/80116702-94654c80-8586-11ea-862f-67f9e69f2647.png)

This is problematic especially as it is difficult to access these field reports on the backend for editing/checking without a title field to click for access.

![image](https://user-images.githubusercontent.com/52411237/80116401-359fd300-8586-11ea-8b14-8e6852487057.png)

URL to the test report on staging: https://dsgoproxyapp.azurewebsites.net/reports/9540

I suggest enforcing the mandatory aspect of this field or alternatively for field reports linked to an emergency but without title to just copy the name of the emergency for title.

Would be great to address soon as 5 minutes after I had by chance ran into this problem on staging while doing some other tests I saw the first instance of this in production.

batpad:
@LukeCaley removing the `backlog` label from this as I think this will be good to fix as part of Field Report validations and cleanups. Also moved to 4.3 board.

nanometrenat:
Agreed - with the new reordering of fields (swapping "search for existing emergency" and "title") there are several being submitted with no title

teklal:
These field reports will then also appear without a link in the weekly digest email
![image](https://user-images.githubusercontent.com/52411237/84006372-e7e0fd80-a96e-11ea-84b1-ba28e6b48585.png)

batpad:
@teklal realized I need some clarification here:

So the behaviour on the field report frontend I believe is desired behaviour - that we wanted users to be able to either link an Emergency or add a title. If we just wish to make the title mandatory always, this is likely the simplest solution.

The other option would be to ensure that the when displaying, if the field report title is empty, it should show the title of the linked emergency. Or a combination of these two things.

Could we please quickly discuss and clarify what the ideal desired behaviour here would be?

batpad:
Thanks for the call @teklal @mmusori - to summarize:

 - We will make `title` always mandatory in the field report, irrespective of whether there is a linked Emergency
[...]

PR #1329, opened 2020-06-15
make title a required field on field reports

batpad:
Refs #1141 - makes the title field on the field report mandatory.


batpad:
Surge URL: https://ifrc-go-fix-1141-fr-title.surge.sh/

PR #1332, opened 2020-06-16
Error message fix for Create Field Report form

GregoryHorvath:
## Changes
`nextProps.fieldReportForm.error` object could have a key like `start_date` if that field has an error, so it wouldn't be good to list all fields here. With this basically it doesn't matter how many fields may have errors it will only display the first and if that's fixed it will show the second (if there is), etc.

This only affects the submission errors, the errors happening at record creation should already say the error was.

cc @teklal


nanometrenat:
@GregoryHorvath please xref the related issue - many cheers

Issue #1125, opened 2020-04-17
[PROD] Field reports: Browser tab not field report specific but whatever GO page the user was previously on


teklal:
## Issue
When clicking on the "Create Field Report" button to start creating a new field report, the text on the browser tab does not change when the field report form opens. Whatever page the user was on before starting to fill the field report, will still stay on the tab.

## Steps to reproduce
Go to any GO page and start filling a field report. Look at what the browser tab says.

A few examples:

![image](https://user-images.githubusercontent.com/52411237/79575229-c166ba80-80c1-11ea-9a38-17581cdafabf.png)
![image](https://user-images.githubusercontent.com/52411237/79575309-dd6a5c00-80c1-11ea-89fb-14c6336578d7.png)
![image](https://user-images.githubusercontent.com/52411237/79575364-f410b300-80c1-11ea-9508-aa356897fe83.png)

....you get the gist.

## Expected behaviour
Tab to say instead "IFRC GO - New Field Report" or "IFRC GO - Field Report" or something like that - can be discussed.

## Related feature
Same issue as with PER tabs #997

batpad:
We need to ensure the page `title` is set correctly on the two pages identified: Create Field Report and PER: while not a huge user impact, I would be highly in favour of fixing this to ensure consistency, and this should be a quick fix. Removing `backlog` label, feel free to add it back, @LukeCaley

LukeCaley:
@teklal can you check that this is working in PROD and close if so. Thanks!

teklal:
Thanks @batpad, for new field reports this is working well. 
Also for PER this works well in most cases. Just when clicking to view already submitted forms (View button on the Account page under the PER forms tab), the tab will still say "IFRC GO - Emergencies".

batpad:
> Also for PER this works well in most cases. Just when clicking to view already submitted forms (View button on the Account page under the PER forms tab), the tab will still say "IFRC GO - Emergencies".

Thanks for the catch on the View Forms @teklal - going to move this back to In Progress - even though it's just this one thing remaining, still part of this issue.

batpad:
Fixed title on the View PER Forms page and pushed to staging.

teklal:
View PER Forms also has the title now say PER, thanks @batpad!

PR #1359, opened 2020-07-02
Fix page title on View PER Form pages

batpad:
Minor, Ref #1125


batpad:
cc @frozenhelium for visibility, minor change to fix title.

Issue #1206, opened 2020-05-06
Highlighted Operations - module cleanup


mariam-yu:
- The title should not be cut but should wrap
- The CHF shouldn't wrap
- YU to provide styling for 4-5 operations (e.g. Asia Pacific region page)

maisa-yu:
Highlighted Operations Modules on pages:
1.0_GO_Home
4.1_GO_Emergencies

imohkay:
@maisa-yu I'm going to work on this over the next couple of days. Just to clarify, will follow the design in Zeplin.
Is the Follow button in red on each operation a new functionality?

imohkay:
@batpad I have pushed to branch feature/1206-highlighted-ops-UI
Could you please checkout and fix the following?

1. The Follow button functionality needs to be implemented. On being clicked, the button should toggle the text "Following" and the style to bounded button (You will need to replace the class 'button--primary-filled' with 'button--primary-bounded'.
2. The translation strings will need to be implemented for the above button's toggle text (i.e. 'Follow' and 'Following')

mariam-yu:
> @maisa-yu I'm going to work on this over the next couple of days. Just to clarify, will follow the design in Zeplin.
> Is the Follow button in red on each operation a new functionality?

@imohkay I'll jump in here. So, it's not a new proposed functionality, but I am not sure at which stage of development it is. This is the option to follow or subscribe to emergencies. @batpad May be you know what's happening with that?

batpad:
> @batpad May be you know what's happening with that?

Yup, so the option to Follow / Subscribe exists on individual emergency pages. We'd need to bring this to all pages that show the Highlighted Operations. This should likely be okay to do, I just need to check-in with @GregoryHorvath and sketch out how it would work, will update here.

imohkay:
@batpad the Highlighted Operations sections needs to come after the Key Numbers on the home page (as per the new designs on Zeplin). But some of the components are one big unit, so I would need your help on this.

batpad:
@mariam-yu - quick question: On the button on the Highlighted Operation cards it says `Follow` / `Unfollow`, whereas on the Emergency page we are currently using `Subscribe` / `Unsubscribe` . Should we change to `Follow / Unfollow` on the Emergency page?

batpad:
@imohkay -

[...]

PR #1362, opened 2020-07-06
#1206: Add new styling for highlighted operations

batpad:
Refs #1206 -

@imohkay just opening this PR for visibility. I will work through the items you've flagged for me on at https://github.com/IFRCGo/go-frontend/issues/1206#issuecomment-653913419

cc @frozenhelium for visibility

### DIFF
--- a/src/root/components/highlighted-operations/operation-card.js
+++ b/src/root/components/highlighted-operations/operation-card.js
@@ -6,10 +6,19 @@ import { formatDate, percent, round, commaSeparatedNumber as n } from '#utils/fo
 import Progress from './../progress-labeled';
 import Translate from '#components/Translate';
 
-const OperationCard = ({operationId, operationName, emergencyDeployments, appeals, lastUpdate, showFollow, isFollowing}) => {
+const OperationCard = ({operationId, operationName, emergencyDeployments, appeals, lastUpdate, showFollow, isFollowing, followOperation, unfollowOperation}) => {
   const beneficiaries = appeals.reduce((acc, curr) => acc + curr.num_beneficiaries, 0);
   const requested = appeals.reduce((acc, curr) => acc + Number(curr.amount_requested), 0);
   const funded = appeals.reduce((acc, curr) => acc + Number(curr.amount_funded), 0);
+
+  function toggleFollow (event) {
+    event.preventDefault();
+    if (isFollowing) {
+      unfollowOperation(operationId);
+    } else {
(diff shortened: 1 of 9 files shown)


src/root/components/highlighted-operations/operation-card.js
```diff
@@ -14,13 +14,18 @@ const OperationCard = ({operationId, operationName, emergencyDeployments, appeal
...
-        <div className="card_box card_box_left">
-          <h2 className='card__title'>{ operationName.length > 30 ? operationName.slice(0, 30) + '...' : operationName }</h2>
-          <small className='last_updated'>
-            <Translate stringId='operationCardLastUpdated'/>
-            &nbsp;
-            {formatDate(lastUpdate)}
-          </small>
+        <div className="card_box card_box_left row">
+          <div className='card__title__wrap'>
+            <h2 className='card__title'>{ operationName.length > 30 ? operationName.slice(0, 30) + '...' : operationName }</h2>
```
frozenhelium, line 19: Not that these are change in this PR but just wanted to point out one thing here:
We should use CSS to properly truncate the text, 
```css
.card__title {
  white-space: nowrap;
  overflow: hidden;
  text-overflow: ellipsis;
}
```
for more info: https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow

imohkay: Okay. Is that a requirement that the title is only one line? Then we can do the needful @frozenhelium 
I suppose we can also make a mixin and reuse as needed...

batpad: @imohkay right now the javascript is doing the truncation to 30 characters in the `.slice(...)` operation - I think the idea here would be to not truncate in Javascript and have CSS handle the display with ellipsis for overflow. Happy to help with the JS end of this if you can implement the text-overflow: ellipsis bit to look neat, etc.

batpad:
@frozenhelium can I ask you to please review my updates to the branch? The implementation of the Follow / Unfollow logic seemed to get a bit strange - let me know if it helps to talk through, I feel like my React skills are getting rusty.

src/root/components/highlighted-operations/operation-card.js
batpad: Per voice with @imohkay - let's ticket and do this separately.

batpad:
This has all been combined into #1369 - closing here.

Issue #1380, opened 2020-07-10
Update header and footer to match Zeplin Designs


imohkay:

imohkay:
@batpad Is the mobile header missing the language select? Or maybe I've missed it?

imohkay:
Done.

Relates to #1364 and matching new designs to what's on Zeplin

PR #1381, opened 2020-07-10
#1380: Improve styling on header and footer, search bar and language …

imohkay:
Matched styles of footer (only font size reduction)
Language dropdown style update and moved to new place
Updated select styles for global search (border radius and borders)


batpad:
cc @frozenhelium for visibility

PR #1383, opened 2020-07-10
Feature/1382 grid ui

imohkay:
Here's an example of the new grid. It has only been applied to the home page, header and footer.

The grid.scss partial in core contains the new styles. I have removed several styles which are creating issues across the site (the nested styles for the divs with class "inner", for example).

The grid uses a container wrapper, with rows and 12 columns with padding for gutters.

Unlike the current site, the gutters' width will remain constant at various screen sizes. The overall width of the site also appears larger on desktop. Overall, the mobile styles have been somewhat improved (but not fully).

Please check the homepage. Other pages may appear broken until the new classes are applied.

This is just an example of how the new grid could work and will need to be applied across the site as we move ahead.


imohkay:
@frozenhelium @batpad PR in draft for now.
@frozenhelium suggestions and feedback are most welcome.
Do let me know if this grid would work for RTL styling.

frozenhelium:
@imohkay This looks fantastic, syntax and usage wise, I haven't ran it yet but it is looking really nice. Gread job :+1:

imohkay:
@frozenhelium good to know. your blessings are very important for this grid :)
One question: 
1. if both margin-left: auto and margin-right: auto are applied on a selector, do I still need to use margin-inline-start and margin-inline-end?

imohkay:
@frozenhelium Hey, I heard that you guys are working on some part of the country 3W page?
I'm planning to migrate most of the pages over to the new grid in the next few days.
I was wondering whether to work on the country page refactoring after your work is done in your branch? Any thoughts? We can do a quick chat if needed...

frozenhelium:
@imohkay If we have same value for left and right (i.e. auto), we don't have to use inline-start or inline-end we can just use left and right.
Also, you can go ahead and migrate to new grid, I'm currently working in the country profile page, but you can work on it anyway. I'll rebase it and resolve conflicts (if any) after your work is complete.
Cheers and Thanks!

imohkay:
@frozenhelium Great :) Thank you :)

Issue #1376, opened 2020-07-09
Home Page: Move Highlighted Ops section below key numbers


imohkay:
Refer to V2 Home Page design

PR #1391, opened 2020-07-14
Move highlighted ops below Key Figures on home-page

batpad:
Addresses #1376 

@imohkay moved highlighted ops below key figures and also handles hiding Highlighted Ops on full-screen so it doesn't interfere with Presentation mode.

If this looks okay to you, let's merge to develop.


imohkay:
@batpad It looks fine, but the Zeplin designs have a full-width edge-to-edge white background. We can implement that in the next release because it will take very long to get it right with the current setup.

PR #1400, opened 2020-07-15
Increase font size for highlighted ops

imohkay:
Increase font size in highlighted ops card
Revery key numbers to previous width and margin

### DIFF
--- a/src/root/components/highlighted-operations/operation-card.js
+++ b/src/root/components/highlighted-operations/operation-card.js
@@ -85,7 +85,7 @@ const OperationCard = ({operationId, operationName, emergencyDeployments, appeal
           <Progress value={requested ? percent(funded, requested) : percent(0.1, 10)} max={100} />
           <div className='card_box_full card_box_container card_box_container--op'>
             <div className="heading-tiny">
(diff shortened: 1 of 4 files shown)


src/root/components/highlighted-operations/operation-card.js
```diff
@@ -67,9 +67,9 @@ const OperationCard = ({operationId, operationName, emergencyDeployments, appeal
 
         <div className='card_box_container card_box_container--op'>
           <div className='card_box card_box_left card_box--op'>
-            <div className="card_box_no">{requested && requested !== 0 ? `${n(requested)} CHF` : '--'}</div>
+            <div className="card_box_no">{requested && requested !== 0 ? `${n(requested)}` : '--'}</div>
```
batpad, line 70: @imohkay - is the removing of `CHF` here intentional?

imohkay: @batpad So the only reason I did it was because the new font sizes were moving text to the next line, and in the Zeplin designs, I realised that the CHF is on the next line anyway. I can easily revert if not desired.

Issue #1386, opened 2020-07-14
Setup "init" pattern with loading to fetch initialization data from backend


batpad:
Relates to #940 

As we move a lot of hard-coded values on the frontend to being handled by the backend as part of #938, we should develop a clean pattern for the frontend fetching data that is needed across most pages. This ensures we have a single place to handle the data fetching and loading state, and will make adding new constants and parameters to fetch from the backend much easier in the future, and avoid the pattern we are using in places currently of hard-coding those values on the frontend.

Data that frontend needs to fetch as part of "init":

 - User data: User logged in state, and profile if the user is logged in
 - Country data
 - ... Ideally we would start moving many constants that are defined both on the back and frontend, but this can be done incrementally once we have this pattern in place.

### Frontend:

 - We would need some way to have the application make an `init` request on load, show a loading state to the user while the init data loads, and only then render the app, so we can be sure the init data is always available across pages once the app has loaded. @frozenhelium I was looking at https://github.com/IFRCGo/go-frontend/blob/develop/src/root/index.js#L101 and realized I don't quite know the best way to do this - might pick your brain here / if you have any thoughts or suggestions, would be very useful.

 - Wire up frontend to fetch data from user `profile` and `init` endpoints.

 - In all the places in the code that currently load hard-coded strings in the frontend code, replace it with getting data from the redux store that has been populated on init.

### Backend:

 - On the backend, we need to decide whether to create a single `init` endpoint, or have the frontend fetch data from different endpoints. My suggestion here would be to have the frontend fetch data from separate end-points for existing entities, eg. countries.
 - We only create a separate end-point for hard-coded constants, not database values (which continue to be served from existing APIs).

So, initially, ideally we would need no change on the backend until we start moving over constants in addition to hard-coded database values. I would rather not prematurely optimize the number of requests we need to make to "init" - if we feel strongly like we need to combine all init values into a single endpoint for performance reasons, we can do this down the line, but this is likely better handled through smart caching strategies, etc.

@frozenhelium - to setup the "init" / loading pattern on the frontend, it would be great to get your help. Let's chat quickly when you have a moment and see how best to lay that out. Then we're happy to jump on integrating the new geospatial APIs onto the frontend and replacing the hard-coded values. Thanks!

cc @geohacker @thenav56

frozenhelium:
@batpad While setting up this I also noticed that we don't have any code splitting (i.e. loading only the requested views / components dynamically) in the GO! We should also implement that in the future.

[...]

PR #1404, opened 2020-07-16
Create a framework to load initial data for app

frozenhelium:
Related to: https://github.com/IFRCGo/go-frontend/issues/1386

### DIFF
--- a/src/root/Multiplexer/index.js
+++ b/src/root/Multiplexer/index.js
@@ -1,0 +1,123 @@
+import React from 'react';
+import { connect } from 'react-redux';
+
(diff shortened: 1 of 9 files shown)


batpad:
cc @kamicut - this is to put in place a pattern to load initial data from the API when the app loads.

Again, this looks good to me, just checking in case you might have any ideas + just for visibility since we likely need to use this when we are loading countries data, etc. from the API, moving over from the hard-coded values on the frontend.

batpad:
@frozenhelium is this okay to hold off on merging until this release is done? All looks great to me, just would rather merge after this release, if that's okay.

frozenhelium:
Yeah, lets hold off till the release

PR #1421, opened 2020-07-21
Feature/exportable view

frozenhelium:


src/root/components/ExportableView/index.js
```diff
@@ -0,0 +1,66 @@
...
+    if (containerRef.current) {
+      // this is to skip current js event loop
+      html2canvas(containerRef.current, {useCORS: true}).then((renderedCanvas) => {
+        Object.values(eventListeners).forEach(listener => {
+          listener(false);
+        });
+
+        startDownload(
+          renderedCanvas,
+          `export-${new Date().getTime()}.png`,
```
batpad, line 29: @frozenhelium it would be nice to be able to customize the filename for the exported file depending on context - is it possible to make the filename a prop (always appending datetime to the string sounds good).

batpad:
@kamicut adding you as reviewer as I'm not really familiar with the Hooks pattern, in case you have any comments or suggestions.

Overall this looks really good to me. @frozenhelium could you link to a ticket in the PR for some context on the issue? thanks!

src/root/components/ExportableView/index.js
frozenhelium: Sure, will do that

Issue #1370, opened 2020-07-08
Change Subscribe / Unsubscribe to Follow / Unfollow on Emergency pages


batpad:
Ref #1206 - for consistency, we should rename Subscribe / Unsubscribe to Follow and Unfollow on the Emergency pages.

@GregoryHorvath assigning to you since you had worked on this, but feel free to toss it back to me.

nanometrenat:
Confirmed that wording is now "Follow" in the surge testing URL https://ifrc-go-release-v4-3-11.surge.sh/emergencies/4059#details

PR #1425, opened 2020-07-21
Change (Un)subscribe to (Un)follow

GregoryHorvath:
Ref.: https://github.com/IFRCGo/go-frontend/issues/1370

Wanted to adjust the related variable names as well, but after checking, even the API calls it "subscribe" so in the end I only changed the text here.


batpad:
Thanks!

PR #1435, opened 2020-07-22
Removed duplicate getAppealsList call

GregoryHorvath:
Ref.: https://github.com/IFRCGo/go-frontend/issues/1443

The home page fired some duplicate requests. The `AppealsTable` component already has the necessary action:
https://github.com/IFRCGo/go-frontend/blob/22407768ad312dfbc51546ecc978ce7ba87f276e/src/root/components/connected/appeals-table.js#L54


batpad:
@GregoryHorvath is it okay to hold off on merging this (and perhaps other optimizations to prevent re-renders / extra data loading, etc) and do this after our next release, so we can keep `develop` relatively stable over the next few days?

PR #1436, opened 2020-07-22
Feature/1434 chf progress bar table headings: Pending changes

imohkay:
@batpad Could we not show the text CHF is there is no value in the table? Also, please could you take a look and see if all is okay? I'm worried I've missed something


batpad:
@imohkay - pushed a fix for when `CHF` would display even when the values were empty. Can you check, and then this should be fine to merge.

PR #1442, opened 2020-07-23
Separate the Map from AppealsTable (stop map re-render)

GregoryHorvath:
Ref.: https://github.com/IFRCGo/go-frontend/issues/1443

## Surge (test) URL
https://ifrc-go-fix-map-rerender.surge.sh/

## Changes
Now when you change pages in an AppealsTable table it won't re-render the Map.

- Added https://github.com/welldone-software/why-did-you-render, made it disabled by default so it doesn't spam the console when not needed
- Removed the `MainMap` component from `AppealsTable`
- Added the `MainMap` component to the relevant views/components (`PresentationDash`, `Region`)
- For the "View All Operations" and similar pages where it wouldn't have the map, added a conditional statement if there is a `title` passed to the `AppealsTable` component it would be wrapped inside a `<Fold>` and having its header changed, while otherwise this would be handled in the parent component

@batpad @frozenhelium if any part needs more explanation or clarification we can do a screenshare anytime


src/root/views/region.js
```diff
@@ -245,21 +252,32 @@ class AdminArea extends SFPComponent {
                         <KeyFiguresHeader fullscreen={this.state.fullscreen} appealsListStats={this.props.appealsListStats} />
                       ) : null}
                       <div className={c('inner', {'appeals--fullscreen': this.state.fullscreen})}>
-                        <AppealsTable
-                          title={strings.regionAppealsTableTitle}
-                          region={getRegionId(this.props.match.params.id)}
-                          regionOperations={this.props.appealStats}
-                          mapBoundingBox={mapBoundingBox}
-                          mapLayers={[this.state.maskLayer]}
```
GregoryHorvath, line 253: Not sure what this did, but I can't find it being used or passed down anywhere.

GregoryHorvath, line 253: Not sure what this did, but I can't find it being used or passed down anywhere. Tried to pass it to the `<MainMap>` component as `layers` prop, but it gave a couple of errors.

@frozenhelium @batpad

batpad:
This is looking really good to me @GregoryHorvath, thanks for doing this!

Could you create a ticket for this, like "Improve performance and reduce data loads on home-page" or so, link this PR, and then add some notes for @karitotp on what to test on the Surge URL? I will take a better look at the code as well.

GregoryHorvath:
>This is looking really good to me @GregoryHorvath, thanks for doing this!
>
>Could you create a ticket for this, like "Improve performance and reduce data loads on home-page" or so, link this PR, and then add some notes for @karitotp on what to test on the Surge URL? I will take a better look at the code as well.

Done. And thanks for the review, I still would liked it more if I could figure out what exactly is getting updated for the `MainMap` component and then maybe a simple `shouldComponentUpdate` would've solved it, I just didn't want to waste more time on this one since it's not really a big upgrade.

PR #1444, opened 2020-07-23
Feature/1382 rtl fixes

imohkay:
This PR is for the following issues on RTL:
- Navigation search bar icon position
- Map filters dropdown arrows position
- The fold link is not flush to left in RTL
- Minor cosmetic fix for Emergencies chart (light grey line between text and progress bar)
- Utility classes for spacing and margin now use inline-start and end.


imohkay:
@frozenhelium These will only show up once the attribute for dir, i.e. rtl is conditionally set on the body, which I'm guessing you will do?

batpad:
@imohkay is @frozenhelium making this change a prerequisite to merging this, or should we merge this to develop? Thanks.

frozenhelium:
@imohkay Yep, I'll do that.
@batpad We can merge these into develop. This should not affect anything unless we have rtl styling.

PR #1453, opened 2020-07-24
add correct img for people targeted and not a font icon

imohkay:


imohkay:
This should fix the issue with the correct icon... @batpad Should be a fast merge :)

PR #1467, opened 2020-07-27
Fix/emergency map heading

imohkay:
Added the new heading Affected Regions to emergency map with new fold markup


imohkay:
@batpad I've just added the fold markup, but haven't included it as a component. Can possibly do this in the next round. Do take a look before merging.

Issue #1189, opened 2020-05-04
[Staging] Fix validation for Field Report form


karitotp:
## Issue

Need to add validators in the date entries and fix the validation for numeric fields to allow entering only positive integer values.

## Steps to reproduce

- Entry the "start date" and "date of data" with upcoming dates.
- Entry decimal numbers on numeric fields.

## Expected behaviour
- The field of dates must not allow the entry of later dates.
- The field of numbers must allow only the entry of integer positive.

## Criticality/Urgency
It is not a blocker, but it should be fixed to have consistent data.

necoline:
@karitotp I am unclear on the ideal validation for the start date. If it's an event, users should select a date in the past, if it's early warning it should be a date in the future right?

karitotp:
Hi @necoline, I assumed that the dates should be only dates in the past, because the date of the data is talking about when was reported the registered and dead cases, and the start date have as help text "The date when the first case is confirmed.". 
I do not sure about a FR as an early warning. I think @nanometrenat can clarify which is the best validation for these dates.

batpad:
Have commented on the PR: but my suggestion for the dates would be that validating that something "is a date" is good enough and let's not get into enforcing specific date ranges or past / future restrictions. I think this can likely get messy with various edge cases, and has the potential to frustrate users a bit. My suggestion would be to just validate that something "is a valid date" and not enforce before / after restrictions.

However, of course, @nanometrenat or @mmusori is likely the best to take a call on this, whether we do want to enforce certain date fields to always be either in the past or the future.

nanometrenat:
Thanks for tagging me in @karitotp @batpad I'm sorry I didn't see this ticket earlier else would've commented.

Agree that we only want to just check that the input "is a date" and not also check past/future.  Partly because the business rules aren't well defined - would not only need clear business rules for early warning vs event when creating the FR, but would ALSO would need to cater for the fact that people often _edit_ existing Field Reports after several days/weeks have passed.

And, being "from the future" myself (NZ being 12-13 hours ahead of UTC) we would also need to ensure that "today" is the right "today" (server time vs user local time) as I know for some systems NZers can only use them in our afternoons (as by that time the Europe date is the same date as NZ).

SO... I suggest keeping with @batpad suggestion and if validation rules are needed then the business (i.e. @mmusori ) can define them once back from leave.  CHEERS ALL

karitotp:
@necoline @batpad, validations are working good.

[...]

PR #1475, opened 2020-07-28
Fix/1189 validation

necoline:
## Issue
#1189 

## Description
This adds additional validation to start date and changes numbers to integers

I also fixed an issue I encountered where I couldn't clear the country dropdown because `updateDistricts` was expecting a value from countries

This resolves everything on the ticket with the exception of the date validation. See issue for further conversation.

### Testing Instructions
https://ifrc-go-fix-1189-validation.surge.sh/reports/new

should error on:
- adding an invalid start date
- adding a number in step 2 (or others) with decimal 

should pass on:
- adding a valid start date
- adding a country, clearing the country and adding again
- adding an integer for fields in step 2 (or others)


batpad:
@necoline this looks good to me!

My suggestion for right now would be that validating that something "is a date" is good enough. I'd be wary of enforcing before / after validations in the dates as I know there likely can be various edge cases where people may want to enter dates in the past / future.

@necoline can we publish this to a surge URL for testing? Instructions in the README / happy to chat about it.

Thanks so much! This is looking great.

necoline:
I don't know how I forgot about my favorite addition to the testing process 🤦 .
https://ifrc-go-fix-1189-validation.surge.sh/reports/new

batpad:
@karitotp could you please test this above URL for the validation fixes and leave comments on https://github.com/IFRCGo/go-frontend/issues/1189 ? thanks!

necoline:
@batpad Karito found an issue with the error messages on the epi-source inputs which I fixed above. 

I think it's ready for a final review

Issue #1108, opened 2020-04-15
Field Report: URLs shared in field report do not show as links on pages generated from field report


teklal:
## Issue
If a user adds a URL in the "Situational Overview"/"Description" field when filling and submitting a field report through the frontend, the URL does not appear as a link on the page generated from the field report even though the "Situational Overview"/"Description" field has a rich-text editor in the backend. Adding a URL (as a link) through the backend to this field does come through correctly, but it is practically never or very seldom that users modify field reports through the backend.

Field with URL in field report
![image](https://user-images.githubusercontent.com/52411237/79330241-7d38b600-7f19-11ea-9a94-f05e67dc8faa.png)

URL on page generated from field report
![image](https://user-images.githubusercontent.com/52411237/79330291-9477a380-7f19-11ea-8338-41c825ae5f51.png)

URL on the emergency page created from the field report
![image](https://user-images.githubusercontent.com/52411237/79330415-c852c900-7f19-11ea-9702-b130d888f7e2.png)

backend of field report
![image](https://user-images.githubusercontent.com/52411237/79330601-1a93ea00-7f1a-11ea-9423-26fdf41f2663.png)

## Expected behaviour
URLs shared through the frontend fields should show up as links in the field report and emergency pages generated from the field report - especially because these respective fields have RTEs enabling the sharing of links in the backend

## Related feature
#1011 
#931

teklal:
This would be great to have addressed as in the new COVID FR we are even asking users to use URLs in field report fields but these do not show as links on the frontend
![image](https://user-images.githubusercontent.com/52411237/80979203-56d6ae00-8e27-11ea-9f13-fbfec3b83e7d.png)

batpad:
This can be tested at https://ifrc-go-fix-1108-description-url-field-report.surge.sh/reports/9432

The only problem is that if a URL is added via the Rich Text Editor in the backend, it will show up with HTML tags in the frontend, but this would be an existing issue with or without this change, some discussion over at https://github.com/IFRCGo/go-frontend/pull/1476

If adding a URL to the field report form on the frontend causes URLs to be correctly linked, I would be in favour of moving ahead here.

@karitotp could you please test that adding a URL in these fields correctly makes them links on the field report detail page?

karitotp:
@batpad, the Urls added in Source details and Situational Overview fields, both are displayed as links in the page.
[...]

PR #1476, opened 2020-07-28
Fix/1108 description url field report

necoline:
#1108 

This fixes the issue of adding a link in 'Situational Overview/Description' and 'Source Details'
Noting from the issue linked above, if you add a link on the RTE on the backend, the frontend will receive the full markup in the form input 

For example a report like this:
image

Will display this in the form:
image

If a user adds a link in the backend, it will likewise display the a tags in the form input field.

I don't know if this is the ideal behavior. 

I implemented a utility to attach a tags to urls in the input fields so they will be successfully displayed in the report as links.

I doubt this is the optimum way to interact with the RTE on the backend, I would love a second opinion on this workaround. Should this url detection happen on the backend? or do we need a better text editor on the frontend?


batpad:
@necoline - thank you for catching the bit about the RTE on the backend, sorry I missed that when scoping this out.

I would still go ahead with this implementation as I see it as an improvement - as mentioned in the other ticket, Field Reports is not something people often edit on the backend, and having links showing up clickable for data entered in the frontend form is valuable. And your code deals with existing links wrapped in `<a>` tags gracefully, so we shouldn't run into situations where we mess-up links that already exist in `<a>` tags.

Whilst saying this, I do think there should be a better medium-long term solution to this. My hunch is that we may need to implement an RTE on the Field Report form at some point. I am still torn on this being the best approach, but if we are going to have the same field RTE editable on the backend, this is probably the right thing to do. @necoline let's catch up and see how we feel about implementing an RTE on the frontend and if this seems realistic to try and do right now. Otherwise, this current implementation looks good to me.

necoline:
Sounds good, thanks for the review @batpad. I think it will be helpful to scope out alternate approaches and identify its priority in the refactoring. 

Here is the surge url I missed before https://ifrc-go-fix-1108-description-url-field-report.surge.sh/reports/9432

batpad:
@necoline I think we should merge this to develop and then flag for further feedback. This looks good to me.

Do you mind rebasing / merging develop here and then we can go ahead and merge. Thanks much!

PR #1495, opened 2020-08-05
Fix global loading not being dismissed in emergency and deployments page

frozenhelium:


batpad:
@frozenhelium I'm still seeing issues on the Region pages.

steps to reproduce:

Navigate to this URL directly, locally: http://localhost:3000/regions/2#operations - loading does not go away.

@frozenhelium interested in having a conversation on what you feel a better more robust fix for this global loading pattern we have, that I can see can be quite prone to bugs like this.

nanometrenat:
> @frozenhelium interested in having a conversation on what you feel a better more robust fix for this global loading pattern we have, that I can see can be quite prone to bugs like this.

I believe this is the refactor proposal:  https://github.com/IFRCGo/go-frontend/issues/1530

Issue #1505, opened 2020-08-07
[PROD] Points on emergency map not appearing until clicked


LukeCaley:
## Issue
The points on the Emergencies map (https://go.ifrc.org/emergencies) do not appear until the `Scale points by` radio button is clicked.

![emergencies-map-error](https://user-images.githubusercontent.com/7791954/89661322-d2d30d00-d8d2-11ea-9371-1a1774143e44.JPG)

Once radio buttons are used the points reappear
![emergencies-map-error-fix](https://user-images.githubusercontent.com/7791954/89661387-ebdbbe00-d8d2-11ea-8de0-e9c1d85dab0d.JPG)
 
BUT whenever the page is reloaded the issue recurs.

## Related feature
Noted during Release v4.3.11 deploy https://github.com/IFRCGo/go-frontend/pull/1465

nanometrenat:
Thanks @batpad @geohacker @GregoryHorvath for deploying the hotfix to Production so quickly.  @LukeCaley to close when happy.

Note for the future when testing in staging - orange dots only appear when DREFs/Appeals are linked to emergencies - i.e. data is needed to check for the orange ones.

PR #1506, opened 2020-08-07
fix points on emergencies page map not showing up

batpad:
Hotfix for broken map on emergency page.

cc @GregoryHorvath @geohacker


nanometrenat:
The ticket for this was https://github.com/IFRCGo/go-frontend/issues/1505#event-3633707045

Issue #987, opened 2020-03-04
[PROD] Field Report won't submit if source selected but no figure populated


nanometrenat:
## Issue
I can't replicate this in Staging however saw for myself that the problem was happening in Production, when a colleague was trying to submit a field report last night.

User was getting the error "Error: Could not submit field report" and could not figure out why - country was valid and all data looked valid.

I looked in user's console and could see that create_field_report was returning a 400 error, with the message {other_num_displaced: ["A valid integer is required"!}

Looking back through the screens of the user's in-progress field report I could see that the user had selected a "source" radio button for the Affected on the Situation screen, but had not input a number.  As the user cannot de-select a radio button once they have selected one, they must put in a figure (e.g. 0) or they cannot submit.

Possible fixes could be:

- expose part of the error message to the user, to allow them to figure out what is wrong, and/or
- strip out the "source" from the API call if the user has selected a source but associated number field is empty.

## Steps to reproduce
![image](https://user-images.githubusercontent.com/33625192/75920009-d9aaaf00-5e5e-11ea-932a-4896d8136193.png)

## Related feature
Several other field report tickets open but don't think related.

## Impact
**Low incidence**

## Severity
Users cannot submit field report if they hit this, and don't know why

## Is there a workaround?
Can put a dummy zero in (if they know to do that)

nanometrenat:
Note that (strangely) the other way is fine - but shouldn't be!  By that I mean, there is nothing to stop me submitting figures but without selecting a source.  If I do that I can submit the FR but the figures I've input just completely disappear.

Suggest also adding validation such that if the user inputs a figure then they MUST select a source.

mmusori:
In addition Health would like the following logic for Numeric details on the epi Field Report:
EPI Field Report Form design (page 2)
1.	“Case” sub-classifications: “Suspect cases”, “Probable cases” and “Confirmed cases” are not independent variables, but rather part of a multi-level classification of “Cases”.  Recommend indenting the “Suspect”, “Probable”, and “Confirmed” terms (within the form design) to show the relationship between the value fields (screenshot below) 

![image](https://user-images.githubusercontent.com/7780809/80381290-09939300-88a1-11ea-8290-76714b368082.png)

[...]

PR #1513, opened 2020-08-10
Enhance/987 epi form

necoline:
#987 
- Added validation to prevent users from adding a source without estimate or vice versa
- Changed estimate input to take numbers to match validation structure
- Added error message to source selection 
- Clears district when country is cleared
- Resolved some console errors

https://ifrc-go-enhance-987-epi-form.surge.sh/reports/new


batpad:
@necoline this looks great!

One observation:

 - Now if I fill in an Estimation but do not specify a Source, it correctly gives me a validation error, which is great. I can now either select the correct Source, or clear the value from the Estimation field to proceed.
 - If I select a `Source` but do not fill in an Estimation, it gives me a validation error. This also seems fine. Except, at this point, the user has no way to "Unselect" the Source and the only way to proceed is to either enter a value or restart the form. I.e. there is no way to "undo" my selecting of a Source and choose to not fill in that data any more.

Not sure what the best solution here is. I think the options would be:

 - Allow a user to "Unselect" an option on the Radio button by clicking it again. This seems a bit strange and not the way I see radio buttons operating ideally and not sure how feasible this is.
- If a user selects a Source but does not fill in a value for Estimation, do not give the user a validation error, and do not send any value for this Source + Estimation to the backend.

Happy to quickly discuss what seems like the best option here. I do imagine having no way to go back on selecting a Source and then being forced to enter a value there or restart the form might be frustrating for users.

batpad:
@necoline - speaking to @nanometrenat it would likely be best to just ignore if the user selects the Source but does not enter a value for the Estimation - i.e. do not show a validation error, and do not submit that data to the backend. Do you think this would be straightforward to implement?

necoline:
My concern with this approach is that the form isn't clearly communicating the field requirements to the user. I'm hesitant to create any presumptive form actions if it's avoidable. In this case, if a user fills in the source but not the estimate, we don't know if they wanted to leave that field blank or if they forgot to add an estimate.

The way I understand it, the estimate/source is a single form field with two values that are required in order to be valid. So clear validation is key in order to communicate this to the user. However, since it is also valid to leave the estimate/source field blank, this option should also be available to the user. However, radio buttons are not designed to be un-selected (as a general rule) so what if we added a third option to clear the estimate/source entry. This way the validation reflects the form rules and allows the user to have complete control over their entries.

I pushed up a commit that displays a `clear` button once a estimate/source entry is made:
image

The validation then gives the user the option of filling in the missing estimate or clearing the source - depending on the case that fits the users needs.
image

necoline:
the surge link is updated with these changes if you want to explore for yourself 🙂

batpad:
I really like this - thank you @necoline for the thoughtful approach here.

@karitotp could you please give https://ifrc-go-enhance-987-epi-form.surge.sh/reports/new a test and test various combinations of field report options (between COVID / non-Covid, Event & Early Warning & Disaster Type = Epidemic options) and test that things seem to work and validation is as expected.

If things seem to functionally work, let's go ahead and merge this in.

karitotp:
@batpad @necoline, in general the validation and the other functionalities of the FR form are working good to me, some things to check below :point_down: 

**In the different field report options:**

- The "clear value" option in some fields is not working,  instead of clear the value the drop-down list is shown.
  - In Disaster Type *
Selection_016

  - In Source (of figures)
Selection_018

- The start date field is mandatory to complete it, but the text of the missing field is not shown as the others.
Selection_017

- Currently selecting the provinces is not mandatory. Is this correct or should this field be mandatory too like the country?
Screenshot from 2020-08-14 00-15-02

- The Source (of figures) is not mandatory neither, even when the numeric details is filled.
Screenshot from 2020-08-14 00-47-21

**In Early warning form:**

- One of the numeric details source is not mandatory as the other sources .
Selection_019

- In planned response the amount fields is not mandatory even when the status options are selected. Same in case of a non-covid event.
Screenshot from 2020-08-14 00-34-16

necoline:
@karitotp Thank you for your thorough review. I added 
- [x] date error message
- [x] error messages for the all EW steps.
- [x] adds the same error handling for planned response

image


I'm not clear on the desired behavior for the source (of figures)

nanometrenat:
> I'm not clear on the desired behavior for the source (of figures)
Hi, this was the proposal in the ticket:  https://github.com/IFRCGo/go-frontend/issues/987#issuecomment-612627196
Many cheers

karitotp:
Hi @necoline, Should I do the testing of this in staging?

batpad:
@karitotp yes, this is all on staging. Thank you!

karitotp:
:heavy_check_mark:  date error message

:heavy_check_mark:  error messages for the all EW steps. 

:heavy_check_mark:  adds the same error handling for planned response

  - The validation here is working good, just need to improve the error message location for a no selected status with the AMOUNT CHF field filled

Selection_025

cc. @necoline @batpad

karitotp:
@necoline, just found that the validation field for only enter integer numbers in Largest Population Centres in Early Warning is not working. This field is allowing letters too.

Screenshot from 2020-08-19 10-15-06

And other thing that may be could be improve is that when the clear button is clicked, that can also clear the error messages.
Screenshot from 2020-08-19 10-06-49

necoline:
@karitotp `Largest Population Centres` is actually meant to be a string. If you look at the description is calls for _names_ of cities. I didn't realize this till I saw a comment for it in the code. I think this point has been missed at a variety of points in the code and I tried to clean it up as much as possible. 

It's good to see the styling of the error message on your screen since it doesn't appear that way for me. I'll do some investigating

PR #1519, opened 2020-08-13
Remove me request from country 3W

frozenhelium:


batpad:
Woops - I know you needed to do the same in your branch @geohacker - just cc'ing you here to keep an eye when rebasing develop.

PR #1521, opened 2020-08-13
Fix/country profile ui

frozenhelium:


src/root/components/header.js
```diff
@@ -148,6 +148,7 @@ function Header (props) {
                 <li>
                   <NavLink to='/' title={strings.headerMenuHomeTooltip} activeClassName='navbar-highlighted' exact>
                     <Translate stringId='headerMenuHome' />
+                    { strings.headerMenuHome }
```
batpad, line 151: This is causing Home to appear twice in the menu :)

batpad:
This looks good to me! (Added a small catch about the Home in the Menu)

Added @imohkay as reviewer for usage of styles, etc.

batpad:
@imohkay you can view this page locally with something like http://localhost:3000/countries/14#operations

src/root/components/header.js
frozenhelium: Oops! Was only meant for testing. Thanks for catching it :+1:

batpad:
@frozenhelium no objections to merging here.

imohkay:
@frozenhelium @batpad this looks okay to me. I haven't compared the 3W page to what's on Zeplin in great detail. Do I need to do that?

frozenhelium:
@imohkay It looks fairly similar. You can quickly check it if you want to

Issue #1535, opened 2020-08-19
[Staging] Editions of FRs cannot be submit


karitotp:
## Issue
The edition of FRs is giving validation error of Planned International Response fields, fields considered in a Early warning event. This is also happening in editions of FRs related to Covid that do not have this fields in the form.

## URL
Edition of a event related to Covid: https://dsgoproxyapp.azurewebsites.net/reports/9643/edit

## Steps to reproduce
- Edit a FR like https://dsgoproxyapp.azurewebsites.net/reports/9643
- Try to submit the edition.
- The page will give the below validation error  

![Screenshot from 2020-08-19 11-26-51](https://user-images.githubusercontent.com/11504548/90666171-9e384d00-e212-11ea-8c49-62b1b25fc3d8.png)

## Expected behaviour
The edition of the FR should be correctly submit.

Issue #1518, opened 2020-08-12
[Staging] ** Cannot continue past page 2 of new Field Report


nanometrenat:
## Issue
"continue" button does not work in New Field Report Situation screen.

## URL
https://dsgoproxyapp.azurewebsites.net/reports/new

## Steps to reproduce
Creating new Field Report in staging

1. I fill in details on page 1 (Context) - I had chosen COVID-19 slider.
2. I click "continue" to go to page 2 (Situation) and page loads fine.
3. HOWEVER when I click "continue" to go to page 3 (Actions) nothing happens - I cannot move forward.  No errors given on screen, just nothing happens.  This occurs regardless of whether I have input any figures or data.

Note that clicking on the direct link to Actions (the red dot below the "Create Field Report" heading) works fine.

Error in console shows the below, not sure if relevant

> index.js:175 Uncaught TypeError: Cannot read property 'split' of undefined
    at a.value (index.js:175)
    at a.value (index.js:226)
    at HTMLUnknownElement.m (react-dom.development.js:188)
    at Object.f (react-dom.development.js:237)
    at y (react-dom.development.js:292)
    at react-dom.development.js:306
    at S (react-dom.development.js:389)
    at react-dom.development.js:414
    at pn (react-dom.development.js:3278)
    at dn (react-dom.development.js:3287)

## Expected behaviour
I should be able to create a Field Report

## Related feature
Potentially related to validation or epi figures changes.

## Criticality/Urgency
Critical as user cannot proceed to create their Field Report.

batpad:
Thanks for the report @nanometrenat 

@necoline assigning this to you if that's okay - the error seems to be coming from here: https://github.com/IFRCGo/go-frontend/blob/develop/src/root/views/field-report-form/index.js#L175 - seems like ensuring description is an empty string there or so should likely fix? But not sure where you best seeing that check happening :-)

Thanks again for the catch @nanometrenat, we have been merging a few things to staging slightly prematurely to not be stuck behind with everything moving, we'll make sure things are thoroughly tested after all the field report fixes / changes are in.

necoline:
I noticed that this morning as well and included the fix in #1513

LukeCaley:
@nanometrenat would you mind doing the UAT honours, please?

PR #1536, opened 2020-08-19
Fix/1535 fr validation

necoline:
#1535
#1513
#1518

The root of the validation issue on the `planned response` is that the initial form values are `undefined` which is what is expected, but once the form is submitted, the field status (the radio button) is given a value of `"0"` which then registers as a "touched" entry. In addition to this, there are a number of fields that are apparently associated with the planned responses even though they aren't displayed (see https://github.com/IFRCGo/go-frontend/blob/b452e22864764286a830d7dc96b1adb9b9e2eee3/src/root/views/field-report-form/data-utils.js#L598)

I created a forced reset of these values on load of an existing form which allows the validation to behave normally. It's not an elegant solution by any means. I'd be happy to talk through a better solution @batpad if you have ideas. 

I also addressed the styling issues here as well and resolved some more form type issues

https://github.com/IFRCGo/go-frontend/pull/new/fix/1535-fr-validation

### DIFF
--- a/src/root/views/field-report-form/cmp-planned-response-row.js
+++ b/src/root/views/field-report-form/cmp-planned-response-row.js
@@ -53,7 +53,7 @@ export default class PlanResponseRow extends React.Component {
               name={`${name}[status]`}
               id={`${name}-status`}
               classLabel='visually-hidden'
-              classWrapper='resp-status col col-8-sm'
+              classWrapper='resp-status col col-7-sm'
               options={options}
               selectedOption={values.status}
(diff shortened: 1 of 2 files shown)


batpad:
@necoline happy to talk this through to see how we might model it better, and we should definitely file a `refactor` ticket around the `plannedResponse` handling. This seems like a reasonable fix for how things are setup currently, going to merge this in. Thanks!

batpad (approved):

PR #1538, opened 2020-08-20
fix useEffect dependency issues, fix lint

batpad:
Fixes an issue where we needed to add all variables used in the function as dependency to the `useEffect` and `useMemo` methods, to get lint / build to pass.

@frozenhelium I'm going to go ahead and merge, but please could you take a look and let me know if this seems correct?

cc @geohacker


frozenhelium:
It looks good.

geohacker:
Thanks for fixing these @batpad!

Issue #1491, opened 2020-08-04
[Staging] Regional Page - and other pages - stuck on loading


nanometrenat:
## Issue
Regional pages on GO are stuck on loading - much of the page loads but then there are the 3 loading dots still going forever.

@mariam-yu raised this first, having tried in an incognito window also - still same issue.  Screenshot below, please investigate.  I will put har file in Teams FYI - the only weird things I can spot in Network tab is that https://dsgocdnapi.azureedge.net/api/v2/featured_event_deployments is returning 301 error and https://dsgoproxyapp.azurewebsites.net/regions/1 is returning 304.  Not sure if either of these explain the issue.

## URL
https://dsgoproxyapp.azurewebsites.net/regions/1#operations

## Steps to reproduce
![image](https://user-images.githubusercontent.com/33625192/89309673-0106e180-d674-11ea-81a7-3ead0c3e484c.png)

## Expected behaviour
Page should load!

## Related feature
Unknown

## Criticality/Urgency
Critical but staging only.

batpad:
Thanks for this catch!

I can reproduce this sometimes on the develop branch as well. @frozenhelium my hunch is it might have to do with the init changes, but I'm not certain. I don't see any obvious errors in the console. I'll dig into this further tomorrow. @frozenhelium, am co-assigning you, but feel free to unassign yourself if you're certain this is not due to the init changes. I need to dig further to be sure, but that's my best hunch currently :)

To be clear, I'm fairly certain this does not affect the release branch at https://ifrc-go-release-v4-3-11.surge.sh/regions/2#operations - can you confirm @nanometrenat ?

nanometrenat:
@batpad confirmed that it does not affect the release branch - thanks for checking :)
xref https://github.com/IFRCGo/go-frontend/issues/1386 as mentioned (whether there's a real link is still being investigated)

nanometrenat:
I am having same symptom on this emergency page in staging:  https://dsgoproxyapp.azurewebsites.net/emergencies/3276#details whereas as at this instant right now the regional page for the Americas (that was getting stuck on loading earlier today) is loading fine.
I tried another emergency (https://dsgoproxyapp.azurewebsites.net/emergencies/3252#details) but am getting a grey screen on that one (i.e. nothing loading at all)

batpad:
Should be fixed with https://github.com/IFRCGo/go-frontend/pull/1495 - thanks @frozenhelium !

batpad:
[...]

PR #1539, opened 2020-08-26
fix stuck loading state on view field report page

batpad:
Fixes the case of global loading not disappearing on https://dsgoproxyapp.azurewebsites.net/reports/9637 when logged in.


nanometrenat:
Xref https://github.com/IFRCGo/go-frontend/issues/1491

PR #1545, opened 2020-09-02
Hopefully this fixes the loading being stuck on Emergency page

GregoryHorvath:
Ref.: https://github.com/IFRCGo/go-frontend/issues/1491#issuecomment-684814928

@frozenhelium I commented out the `syncLoadingAnimation` part because I suspect there is some way to getting `shouldShowLoadingAnimation == true` and `!this.loading == true` at the "last render() call" of the page but I couldn't get to the bottom of that. Also note `this.props.siteRepResponse` doesn't have `fetching` but `this.props.situationReports` has.

### DIFF
--- a/src/root/views/emergency.js
+++ b/src/root/views/emergency.js
@@ -136,9 +136,17 @@ class Emergency extends React.Component {
     if (newProjectAdded) {
       this.setState({ showProjectForm: false });
     }
+
+    if (
+      (nextProps.projectForm.fetched || (!nextProps.projectForm.fetched && !nextProps.projectForm.fetching))
+      && nextProps.event.fetched && nextProps.situationReports.fetched
+    ) {
+      hideGlobalLoading();
+    }
   }
 
   componentDidMount () {
+    showGlobalLoading();
     this.getEvent(this.props.match.params.id);
     this.props._getSitrepTypes();
     if (this.props.isLogged) {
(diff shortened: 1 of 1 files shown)


batpad:
@frozenhelium could you please review?

frozenhelium:
@GregoryHorvath @batpad I'll check it out and get back

frozenhelium:
@GregoryHorvath this current implementation seems to be working. But, @GregoryHorvath  @batpad I think we should avoid second guessing and applying alternative methods in hope that it'll get fixed and use correct method. I've created a proof of concept for new global loading and created a [draft PR](https://github.com/IFRCGo/go-frontend/pull/1549). Let me know what do you guys think.

frozenhelium (approved):

GregoryHorvath:
Closing this in the light of: https://github.com/IFRCGo/go-frontend/pull/1549

batpad:
Am re-opening this since I'm not  certain we will have #1549 ready, and we may need to keep this patch-work around the global loading a bit longer. The approach in #1549 seems great, just not sure if we can apply it to all pages in time for release, and I'd rather not have some pages using the old approach and some the new.

Let's discuss how we feel about the current state of the loading from a user-perspective and if we can afford to delay the refactor of loading.

@GregoryHorvath @frozenhelium

frozenhelium:
Closing again as we're proceeding with : #1549

PR #1549, opened 2020-09-04
Feature/new global loading

frozenhelium:
@batpad  @GregoryHorvath I think we should avoid using the previously used global loading. There are a lot of things that is wrong with it. So, I purpose new way of showing the global loading. It's exactly same as before but the developer API is slightly different. It uses react context to sync state with the component mounted in `app.js`. Please let me know your thoughts.

### DIFF
--- a/src/root/components/NewGlobalLoading/index.js
+++ b/src/root/components/NewGlobalLoading/index.js
@@ -1,0 +1,36 @@
+import React from 'react';
+
+import Spinner from '#components/spinner';
+
+export const GlobalLoadingContext = React.createContext({
+  loading: false,
+  setLoading: () => { console.warn('setLoading on GlobalLoadingContext called without initialization'); },
+});
+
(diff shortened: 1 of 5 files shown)


GregoryHorvath (approved):
I'd lie if I said that I understand any of this, but for sure it solved the loading getting stuck issues for me.

nanometrenat:
Ref.: #1491

batpad:
@frozenhelium would be great for you to give me and @GregoryHorvath a quick explanation on the new React concepts like `useMemo`, `useContext`, etc :-) -- it's taking me a little bit to grapple and fully understand the code here, but it seems to work for me and most definitely the pattern looks like it makes more sense than what we were doing.

@frozenhelium can you say what it would take to finish this and make it non-Draft?

frozenhelium:
@batpad It works good but is currently only implemented in countries and emergency page. We can similarly implement this in other pages as well.

Issue #1550, opened 2020-09-07
[Staging] Key figures on home-page showing Data not Available


batpad:
## Issue

On staging, https://dsgoproxyapp.azurewebsites.net/, I see `Data not available` for the Key Figures on the home-page. I am not fully certain why and whether this is a backend or frontend issue.

## URL

https://dsgoproxyapp.azurewebsites.net/

## Steps to reproduce

Go to URL, expect to see Key Figures at top of home-page, instead see `Data not available`.

## Expected behaviour

I should see Key Figures at the top of the home-page.

## Related feature

I am uncertain whether this is related to backend API changes, or something on the frontend. Need to dig further.
cc @GregoryHorvath @thenav56 @frozenhelium 

## Criticality/Urgency

This is a release blocker.

batpad:
Closing this to reduce clutter.

PR #1551, opened 2020-09-07
Fix homepage key-figures (null error for centroid)

GregoryHorvath:
Ref.: https://github.com/IFRCGo/go-frontend/issues/1550

## Notes
- `console.log()`-ed error: `TypeError: Cannot read property 'coordinates' of null (key-figures-header.js:74)`

Also keep in mind https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#Browser_compatibility
I've ran into issues with this with the Tableau Web Data Connector, because the application uses some older version of the Chromium engine.

@geohacker Feel free to take over if you think this wouldn't fix the whole issue.

### DIFF
--- a/src/root/utils/utils.js
+++ b/src/root/utils/utils.js
@@ -70,7 +70,7 @@ export function aggregateCountryAppeals (appeals, countries) {
         }),
         geometry: {
           type: 'Point',
-          coordinates: getCountryMeta(countryAppeals[0].country.id, countries).centroid.coordinates || [0, 0]
+          coordinates: getCountryMeta(countryAppeals[0].country.id, countries).centroid?.coordinates || [0, 0]
         }
       };
     })


batpad:
This looks good to me :+1: -- @geohacker could you take a look and check if this pattern maybe repeated elsewhere / also if coordinates showing up undefined is some sort of anomaly?

> Also keep in mind https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#Browser_compatibility

Hmm this seems like there should be some `babel` rule to handle in the transpilation step? @frozenhelium do you know off-hand if `babel` should take care of this? I'll take a look as well.

This PR looks fine to merge for me.

batpad (approved):

geohacker:
This is a bit weird that centroid object is missing. It should always be there and the value could be null. But this fix looks good to me. And yes, would be good to add some babel rule to handle compatibility.

geohacker (approved):

PR #1554, opened 2020-09-08
New global loading on all pages

batpad:
@frozenhelium added the new global loading to accounts and deployments pages.

Could you please give this a quick review and if it looks okay, I can proceed with implementing this across all the pages where we currently use the old loading pattern.

Thanks!

cc @GregoryHorvath

### DIFF
--- a/src/root/views/account.js
+++ b/src/root/views/account.js
@@ -33,7 +33,6 @@ import { getCountryMeta } from '#utils/get-country-meta';
 import { countries, disasterType, orgTypes } from '#utils/field-report-constants';
 import { apiPropertyDisplay, apiPropertyValue } from '#utils/format';
 import NewGlobalLoading from '#components/NewGlobalLoading';
-// import { showGlobalLoading, hideGlobalLoading } from '#components/global-loading';
 import { showAlert } from '#components/system-alerts';
 
 import Fold from '#components/fold';
@@ -227,7 +226,6 @@ class Account extends React.Component {
 
   componentDidMount () {
     this.componentIsLoading = true;
-    // showGlobalLoading();
     const { user, _getProfile, _getFieldReportsByUser, _getPerCountries, _getPerDocuments, _getPerDraftDocument } = this.props;
     _getProfile(user.username);
     _getFieldReportsByUser(user.id);
@@ -281,9 +279,6 @@ class Account extends React.Component {
       this.props._getPerDraftDocument(draftQueryFilters);
     }
 
-    // if (nextProps.profile.fetched === true) {
-    //   hideGlobalLoading();
-    // }
   }
 
   syncNotificationState (data) {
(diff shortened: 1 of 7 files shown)


src/root/views/deployments.js
```diff
@@ -267,6 +265,12 @@ class Deployments extends SFPComponent {
 
   render () {
     const { strings } = this.context;
+    const pending = this.isPending(this.props.eruOwners);
```
frozenhelium, line 268: We can directly do
```
const pending = this.props.eruOwners?.fetching
```

src/root/views/deployments.js
```diff
@@ -267,6 +265,12 @@ class Deployments extends SFPComponent {
...
+    if (pending) {
```
frozenhelium, line 269: I think we can directly do
```
<App className='page--deployments'>
    { pending && <NewGlobalLoading /> }
    ...
```

src/root/views/account.js
```diff
@@ -791,6 +792,12 @@ class Account extends React.Component {
 
   render () {
     const { strings } = this.context;
+    const pending = this.isPending(this.props.profile);
+    if (pending) {
+      return (
+        <NewGlobalLoading />
+      );
+    }
```
frozenhelium, line 800: we can figure out pending inline as
```
const pending = this.props.profile?.fetching;
```

And put the global loading without blocking the view i.e. inside `App`
```
<App className='page--account'>
    { pending && <NewGlobalLoading /> }
    ...
```

src/root/views/deployments.js
batpad: So the problem with that was that the rest of the render depends on variables that are not available yet and that causes errors. I could go through on each page and make sure that things don't error, but that seems a bit brittle, and in some cases it seems much safer to short-circuit the return and not attempt to render the main view. Any suggestions here for doing this better? This didn't seem great to me, but could not figure a better way without getting deep into each view to ensure it doesn't error if data isn't available yet.

src/root/views/account.js
batpad: Same as above.

src/root/views/deployments.js
frozenhelium: So for that what we can do is conditionally call the `renderContent()`. Something like
```jsx
<App className='page--deployments'>
    ...
    { this.props.eruOwners?.fetched && this.renderContent() }
```

frozenhelium: The reason we want to render `App` is that it'll show the headers. If we don't it'll look similar to how emergencies page currently load (i.e. without header / navbar)

batpad:
@frozenhelium we did not manage to land this :( - hopefully this can be worked on as part of request and loading refactors, but closing this to clean up.

PR #1562, opened 2020-09-09
show only independent countries in the country dropdown

geohacker:
related to https://github.com/IFRCGo/go-api/pull/883

This PR ensures that the country dropdowns now only show countries that are independent.

### DIFF
--- a/src/root/utils/field-report-constants.js
+++ b/src/root/utils/field-report-constants.js
@@ -54,6 +54,11 @@ export const countries = (countries) => {
     countriesSelectList = countries.map(country => ({ value: country.id, label: country.name }));
   }
 
+  // show only independent countries.
+  countriesSelectList = countriesSelectList.filter(country => {
+    return country.independent;
+  });
+
   return [
     {value: '-- Country --', label: ''},
     ...countriesSelectList,
--- a/src/root/utils/field-report-constants.js
+++ b/src/root/utils/field-report-constants.js
@@ -56,7 +56,14 @@ export const countries = (countries) => {
 
   // show only independent countries.
   countriesSelectList = countriesSelectList.filter(country => {
-    return country.independent;
+    // incase country doesn't have independent property, include anyway
+    if (!country.hasOwnProperty('independent')) {
+      return country;
+    }
+
+    if (country.hasOwnProperty('independent') && country.independent) {
+      return country.independent;
+    }
   });
 
   return [
--- a/src/root/utils/field-report-constants.js
+++ b/src/root/utils/field-report-constants.js
@@ -51,20 +51,15 @@ export const countries = (countries) => {
   if (countries[0].hasOwnProperty('value') && countries[0].hasOwnProperty('label')) {
     countriesSelectList = countries;
   } else {
-    countriesSelectList = countries.map(country => ({ value: country.id, label: country.name }));
+    // show only independent countries (include countries with empty or unavailable 'independent' prop)
+    countriesSelectList = countries
+      .filter(
+        country => country.independent === true
+        || country.independent === undefined
+        || country.independent === null
+      )
+      .map(country => ({ value: country.id, label: country.name }));
   }
-
-  // show only independent countries.
-  countriesSelectList = countriesSelectList.filter(country => {
-    // incase country doesn't have independent property, include anyway
-    if (!country.hasOwnProperty('independent')) {
-      return country;
-    }
-
-    if (country.hasOwnProperty('independent') && country.independent) {
-      return country.independent;
-    }
-  });
 
   return [
     {value: '-- Country --', label: ''},
--- a/src/root/utils/field-report-constants.js
+++ b/src/root/utils/field-report-constants.js
@@ -46,19 +46,22 @@ export const getVisibility = (strings) => [
 
 // This transforms a countries collection to be used
 // directly in a select dropdown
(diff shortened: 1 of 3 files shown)


src/root/utils/field-report-constants.js
```diff
@@ -54,6 +54,11 @@ export const countries = (countries) => {
     countriesSelectList = countries.map(country => ({ value: country.id, label: country.name }));
   }
 
+  // show only independent countries.
+  countriesSelectList = countriesSelectList.filter(country => {
```
GregoryHorvath, line 58: This won't work in the case of:
https://github.com/IFRCGo/go-frontend/blob/edd6de367a66b7eae24ea2757ce161e0bedbcdfd/src/root/utils/field-report-constants.js#L53-L55

For example it doesn't work for me on the Registration page or the Create New Field Report.

Just a suggestion (doesn't need to change), but something like this would be a bit more clear/n for me:
```javascript
countriesSelectList = countriesSelectList.filter(country => country.independent === true);
```

GregoryHorvath:
Waiting on merging this for @jhenshall to confirm if we really want this for pages like the Registration page

GregoryHorvath:
@geohacker Simplified the filtering a bit: https://github.com/IFRCGo/go-frontend/pull/1562/commits/c4186ac2e46935cdcdf9fa64a50ed278a3f544d9#diff-839e092a678fa647520c370aaf1dc57cR55-R60

Sorry for me keep hanging onto this. My change also includes `undefined` and `null` values of `independent` alongside `true`. Is it really possible that some countries might not have this property? We can definitely go with it for now, but is it feasible to have a straight `false` or `true` for all countries, or the `null` is there because for some it might be questionable if this is true?

jhenshall:
Thanks for this @geohacker and @GregoryHorvath ! I'm still looking at edge cases for the site-wide dropdown logic but re: the missing values, you may already have these covered but just a couple of thoughts:

- Is there a difference between `null` and `undefined` in this case?
- An example where this should probably be kept is for clusters and regions - so the filter should be to exclude 'countries' with `false` rather than include those with `true`
- It looks like the data was added based on iso3? Means some have been missed for various reasons but I think this is fine for me to pick up the edge cases in the backend.

GregoryHorvath:
> Thanks for this @geohacker and @GregoryHorvath ! I'm still looking at edge cases for the site-wide dropdown logic but re: the missing values, you may already have these covered but just a couple of thoughts:
> 
> * Is there a difference between `null` and `undefined` in this case?
> * An example where this should probably be kept is for clusters and regions - so the filter should be to exclude 'countries' with `false` rather than include those with `true`
> * It looks like the data was added based on iso3? Means some have been missed for various reasons but I think this is fine for me to pick up the edge cases in the backend.

@jhenshall **TL-DR: we're good with the current code, it's just like me VS `null` values for code-maintainability's sake**

1. this is only a slight inconvenience but for better code maintainability I'd say it would be best if all countries had the property but I haven't fully checked if this was the case already
    - `null` would mean the Country object has the `independent` property but it has no value
    - `undefined` means it doesn't even have the property and these have to be separately handled
2. also this way we need to decide what do we do with these ones which have neither `true` or `false` there, currently we include them alongside the ones with `independent: true` (so we show `true`, `null` and `undefined` ones)
3. it was updated with the admin0 shapefile with the logic which checks if it contains and 'INDEPENDEN' value per country as much as I understood from quickly checking the code

Sorry didn't want this to turn into some war against `null`s, I'm just saying if there is no foresight that we'd ever need `null`s there we could drop that as an option and ensure every Country object has the `independent` property with only a `true` or `false` value. It would benefit both code maintainability and more than ever simplify the logic on which countries to include/exclude.

jhenshall:
@GregoryHorvath - loving the war against the nulls! Great to really think this through and I'm all for simplifying things. However, with regions and cluster offices in the country list, it feels weird to have them as `independent` = `true` and feel that `null` could be more appropriate? Having said that, I think we should treat this as the interim fix before the longer term geo stuff - so happy to be talked out of this if it makes a big difference to the code, and we can revisit later.

geohacker:
@GregoryHorvath thanks for catching these potential issues. I think it's a good idea to guard against `undefined` but that should never be the case. The independent field is a NullBoolean so it will be either false, true or null — but anyway, this is fine. 👍 

By default it'll be `null` unless the shapefile explicitly says true or false. 

I just pushed a fix to make the filter optional. This will now filter only on the field report form page to show countries that are independent but will continue to list all countries on the registration page. I think that should address all the concerns so far on this PR.

GregoryHorvath (approved):

batpad:
This looks good to me. @GregoryHorvath @geohacker hit the merge button?

GregoryHorvath:
@batpad Sorry, I still don't know who can/can't merge besides you and me.

geohacker:
Thank you! I just deployed this on surge in case @jhenshall wants to give this a test before merging https://ifrc-go-show-independent-countries.surge.sh/register -- but also good on my end to merge 👍

batpad:
> Sorry, I still don't know who can/can't merge besides you and me.

I think we can all merge :) - I just didn't want to do this one since I wasn't 100% sure if everything was done and clear. Thanks all!

geohacker:
For example — now on the register page:

image


On the field report form:

image

PR #1564, opened 2020-09-10
Cache response for all countries and regions

frozenhelium:
@batpad Alternative approach for caching countries and regions


batpad:
This seems to work well for me.

@frozenhelium I remembered this module I've used in the past: https://www.npmjs.com/package/redux-persist - not sure why I didn't think of this before. Would totally trust your call here on which you feel is the better approach.

frozenhelium:
@batpad I think, manually caching is better suited for our current case. We have more control over the expiry and other stuff.

batpad:
@frozenhelium extremely weird, but in this branch, the Key Figures on the home-page disappear for me - shows up as Data Not Available (while this is fine in `develop`) - I can't really figure a reason for this and I don't see any errors, so am looking as well, but could you take a look in case you spot something obvious?

cc @GregoryHorvath since I know you looked at a similar issue with Key Figures recently? Thanks!

GregoryHorvath:
> cc @GregoryHorvath since I know you looked at a similar issue with Key Figures recently? Thanks!

@batpad Those happened because of missed nullcheck for the `centroid`s but I see those are handled here.

frozenhelium:
@batpad try clearing out the cache once, you can go to  `/clear-init-cache`

Issue #1548, opened 2020-09-03
[PROD] Grey screen when opening GO in IE - suggest a "use other browser" warning


nanometrenat:
## Issue
Users accessing GO with Internet Explorer just see a completely grey screen and think something is broken.

It has explicitly been decided that GO is not supported on IE; however we need some way of telling the user that so we don't get support calls.

It would be great if GO could detect the user's browser, and display wording such as "Internet Explorer cannot be used to access GO.  Please use an alternative browser"
@LukeCaley @tovari to come up with good words and prioritisation

## Steps to reproduce
Go to any GO page in Internet Explorer

## Expected behaviour
Users should know what to do

## Related feature
Past Browser compatibility tickets include https://github.com/IFRCGo/go-frontend/issues/1537 #772 #921 #767

## Impact
Anyone using IE

## Severity
Completely stops them from using GO

## Is there a workaround?
Install/use a different browser

batpad:
cc @frozenhelium for advice / recommendations on the best approach here.

My hunch / suggestion would be to do browser detection early, and show a simple message to IE users saying we are very sorry, but they are using an unsupported browser and they can use any of these other browsers - @frozenhelium how feasible do you think this is?

Would be good to know how important this is from a business perspective.

Based on conversations with @frozenhelium and all our reading into this, supporting Internet Explorer (which is by now not supported by Microsoft) would hold back development too much and not allow us to use modern browser features. But, we should definitely be better with our messaging and not show users a grey screen. @frozenhelium can you please advise on how much work you see this being, to help prioritize? Thanks!

frozenhelium:
@batpad We can block the further processing / rendering of the GO app for IE.
We need to detect the browser first. [browser-detect](https://www.npmjs.com/package/browser-detect) an excellent library for that. We can then conditionally initialize the React app in the `src/index.js` according to the browsers we support.
We can also show the recommended browser if we detect that user is currently running unsupported browser (i.e. IE)

batpad:
@LukeCaley @mariam-yu could we get copy for text to show the user if they are using Internet Explorer?

@frozenhelium - thanks! So we'd just add some conditional like `if (browserIsNotIE) { ... render the React app .. } else { document.getElementById('app-container').innerHTML = 'Sorry we do not support your browser' }` <-- something like this?

@GregoryHorvath might this be something you can take on once we have some text and minimal design for this message to users?

GregoryHorvath:
[...]

PR #1566, opened 2020-09-10
Check if user's browser is supported

GregoryHorvath:
Ref.: https://github.com/IFRCGo/go-frontend/issues/1548

Not entirely sure if I put all this in the right place, and if I used `useMemo` right (it's for something like caching values?)

## Changes
- Added https://www.npmjs.com/package/browser-detect?activeTab=readme
- The `Root` component's content is only rendered if the browser is not IE or Edge (probably this will be adjusted)

### DIFF
--- a/src/root/index.js
+++ b/src/root/index.js
@@ -1,6 +1,5 @@
 import React from 'react';
 import { Provider } from 'react-redux';
-import browser from 'browser-detect';
 
 import store from '#utils/store';
 import { detectIE } from '#utils/ie';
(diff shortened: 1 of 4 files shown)


src/root/index.js
```diff
@@ -10,22 +11,37 @@ import Multiplexer from './Multiplexer';
 
 require('isomorphic-fetch');
 
+const browserNames = {
+  ie: 'Internet Explorer',
+  edge: 'Edge',
+  firefox: 'Firefox',
+  chrome: 'Chrome'
+};
+const unsupportedBrowsers = ['ie', 'edge'];
```
batpad, line 20: This should only be IE and not Edge?

batpad:
@GregoryHorvath it is likely a better idea to not load the React component at all, since even that may not work in IE - so @frozenhelium suggested putting it in the root index.js here: https://github.com/IFRCGo/go-frontend/blob/develop/src/index.js -- so here we check the browser, and just use a plain document.getElementById('app-container').innerHTML = '' to set the HTML and not React methods. Let me know if that makes sense, cc @frozenhelium

src/root/index.js
GregoryHorvath: https://github.com/IFRCGo/go-frontend/pull/1566/commits/f953f2436c7116616e00271d1b49a862a941b12a moved the handling to [/src/index.js](https://github.com/IFRCGo/go-frontend/blob/develop/src/index.js) only including IE now

batpad:
This looks good to me @GregoryHorvath - do we have a good way to test this actually works in Internet Explorer?

GregoryHorvath:
@batpad Good question, just deployed it to surge (database is not set, doesn't matter here): https://ifrc-go-feature-browser-detect.surge.sh

But IE almost crashes under it and still just showing the grey screen. I'll have to look into this more.

GregoryHorvath:
@batpad Well, I'm somewhat stuck with it with no idea where to go next.

It gives 2 errors:
image

One is pointing to an => function in `/node_modules/ansi-styles/index.js`, and one is to the `Multiplexer`
```javascript
const wrapAnsi16 = (fn, offset) => function () {
	const code = fn.apply(colorConvert, arguments);
	return `\u001B[${code + offset}m`;
};

---------------------------------------------

function Multiplexer(props) {
  const {
    getUser,
```

I've tried:
1. adding IE to the `targets` in Babel config
2. adding `import 'react-app-polyfill/ie11';`
3. removing all other imports from `/src/index.js` than `import browser from 'browser-detect';`

Even with 3. it gave me the error with the => function.

GregoryHorvath:
Closing this, we don't want to use a 3rd party package for this, simple JS should do.

Issue #1541, opened 2020-08-27
[Staging] Register screen stuck on loading


jhenshall:
## Issue
Register page on staging site gets stuck on loading and displays grey screen. In console, says `GET_COUNTRIES_FAILED` so could be linked to geo api changes? 

## URL
https://dsgoproxyapp.azurewebsites.net/register

## Steps to reproduce
Visit URL above.

## Expected behaviour
Page loads.

## Related feature
Possibly the geo api changes #1534 ?

nanometrenat:
@GregoryHorvath this wasn't on the board so I added it to 4.3 so people know where it's at.  Can you please put it to Ready for QA when it's ready?  thanks

batpad:
This should now be fixed in staging. Can you please confirm @jhenshall ?

cc @karitotp

jhenshall:
Thanks! The grey screen issue seems fixed but in trying to create a new account, the country field fails with 'should be string'. Is this a separate issue? Maybe related to the geo work?
cc. @batpad @geohacker @karitotp 

![image](https://user-images.githubusercontent.com/6359748/92917012-fa634b00-f425-11ea-9ea3-12980a87e25c.png)

batpad:
Thanks @jhenshall !

@geohacker do you know what might be the cause of this validation error? Something about the "type" of the country id and that needing to be updated in the validation schema?

cc @GregoryHorvath

geohacker:
Ah yes, this is a validation issue. I'll fix! Thanks for catching @jhenshall!

batpad:
The validation issue should now be fixed in staging.

@jhenshall could you please test and confirm?

@karitotp Would also be good to give a full test to the Register page functionality after this fix. Thank you!

jhenshall:
Thanks @batpad - successfully signed up as new user, so all good.

karitotp:
@batpad, in terms of mandatory fields, seems is not working good for the City an the Organization name when National society is selected, or could be that only there is missing the validation message

![Selection_051](https://user-images.githubusercontent.com/11504548/93097938-d4c78300-f66b-11ea-9a66-ca3d8c12dfe9.png)

batpad:
@karitotp yea, i think here the problem is that it only shows the first validation error and not all when submitting. We can track this as a separate issue if needed, cc @GregoryHorvath . thank you!

jhenshall:
[...]

PR #1572, opened 2020-09-14
Fix/1541 country validation

batpad:
Refs #1541 - country ids are now numbers, not strings.

@geohacker can you check if this seems okay?


batpad:
Oops, this accidentally includes a commit from #1571, should be fine to ignore that, will just merge that before this.

geohacker (approved):

PR #1581, opened 2020-09-16
Multilingual map labels

geohacker:
This PR hides the default text layer on Mapbox maps and introduces a custom layer to show country labels based on the users current language selection.

### DIFF
--- a/src/root/components/connected/emergencies-dash.js
+++ b/src/root/components/connected/emergencies-dash.js
@@ -10,6 +10,7 @@ import EmergenciesMap from '#components/map/emergencies-map';
 import Progress from '#components/progress';
 import BlockLoading from '#components/block-loading';
 import Translate from '#components/Translate';
+import { countriesGeojsonSelector } from '../../selectors';
 
 class EmergenciesDash extends React.Component {
   renderEmergencies () {
@@ -158,7 +159,10 @@ class EmergenciesDash extends React.Component {
               </div>
             </div>
           </div>
-          <EmergenciesMap lastMonth={lastMonth} />
+          <EmergenciesMap
+            lastMonth={lastMonth}
+            countriesGeojson={this.props.countriesGeojson}
+          />
         </section>
       </div>
     );
@@ -177,7 +181,8 @@ if (environment !== 'production') {
 
 const selector = (state) => ({
   lastMonth: state.emergencies.lastMonth,
-  aggregate: state.emergencies.aggregate
+  aggregate: state.emergencies.aggregate,
+  countriesGeojson: countriesGeojsonSelector(state)
 });
 
 export default connect(selector)(EmergenciesDash);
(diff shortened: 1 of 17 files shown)


geohacker:
Deployed on surge to test https://ifrc-go-multilingual-map-labels.surge.sh/

One edge case I discovered is 'State of Palestine' label. This comes as a separate layer from Mapbox and doesn't have a corresponding 'country' in the Go database. @jhenshall could you take a look at this closely? Thanks!

cc @batpad @guipizzini @LukeCaley

batpad:
This is looking really amazing @geohacker !

We should soon have the actual country translations on staging soon so should be able to see this with real translated data instead of the `(fr): `, `(es): ` stuff right now.

I'm a bit curious how Arabic labels display and if it handles RTL gracefully.

This still seems to me to be too big a change to try and get into this release, but if you and @jhenshall can start working out the edge cases and kinks, this seems like it would be fabulous to try and get in soon after the current release - from what I tested and looked at on this PR, the labels seem to work really well.

jhenshall:
Impressive stuff @geohacker, looks great! Nice spot for the 'State of Palestine' label. This should correspond to 'Palestine' in the GO database but aware of the weird extra layer in MapBox - does that complicate things then? Agree with @batpad that this seems a big change to be adding late on but great to have ready for a follow up release.

geohacker:
> This should correspond to 'Palestine' in the GO database but aware of the weird extra layer in MapBox - does that complicate things then?

Oh sweet, then we just need to disable that extra layer coming from Mapbox and use the palestine label from the Go db. I can make that change and push here for us to test.

batpad:
The country translation labels have been updated in the backend, and one can see the map updated in different languages at https://ifrc-go-multilingual-map-labels.surge.sh/

Note: Some countries still have the language prefix like `(es) American Samoa` or so instead of the Spanish translation. We are following up on the mismatch of countries between staging and production and ensuring all countries on production have translations separately - the `(es)` etc prefix given to country names was done just on staging to make functional testing easier before we had actual translations..

I am very tempted to propose getting this in for this release because of how nice it looks to have things translated on the map. It would be great to get feedback on the Arabic handling on the map and that everything looks okay.

@karitotp will also be great to test this:

 - Switch language, and navigate to any map on the site.
 - Country labels should now be in the language you have chosen.

This should apply to all maps across the site.

Let's wait for some testing feedback and revisit the call on whether to merge this in for this release? It is tempting because of how nice the map looks translated, but will let @jhenshall and @geohacker weigh in here once we have testing feedback.

karitotp:
> Country labels should now be in the language you have chosen.

It looks good on the different maps in the site, there are still countries that are not being translated and in their place are the initials of the selected language. On the other hand, the Ivory Coast country remains as `Côte d'Ivoire` in English, Spanish and French ​​except in Arabic.

cc. @batpad @geohacker

geohacker:
This looks really nice overall and I can see how it'd improve the experience around multiple languages, though, I think it would be wise to just wait to merge this until this upcoming release is over. We can likely merge and deploy this right after that. 

In the meantime, we can ensure the geo data on the Go DB is all satisfactory along with accurate translations.
cc @jhenshall @batpad

batpad (approved):

batpad:
Merging this to staging to test.

PR #1587, opened 2020-09-16
Remove border-bottom from Lang. Select style

GregoryHorvath:
Aligns the Language dropdown with the other elements of that line.

### DIFF
--- a/src/root/components/LanguageSelect/styles.module.scss
+++ b/src/root/components/LanguageSelect/styles.module.scss
@@ -18,7 +18,6 @@
   display: flex;
   align-items: center;
   padding: 3px 6px;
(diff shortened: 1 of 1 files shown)


batpad:
@imohkay do you mind taking a quick look and reviewing?

@GregoryHorvath can we remove that line of code instead of commenting out?

GregoryHorvath:
@batpad Sure, should I also create a surge url?

batpad:
> @batpad Sure, should I also create a surge url?

Na, this seems small enough to just be tested in staging. Just want to get @imohkay's blessings for CSS changes.

imohkay:
@batpad This looks good to me.

batpad (approved):

PR #1592, opened 2020-09-17
Get Disaster Types from DB

GregoryHorvath:


src/root/utils/get-dtype-meta.js
```diff
@@ -1,6 +1,10 @@
-import { disasterType } from './field-report-constants';
+import store from './store';
```
GregoryHorvath, line 1: @frozenhelium @batpad Really afraid of this, but couldn't really resolve it any other way without rewriting the whole thing.

src/root/views/countries.js
```diff
@@ -363,7 +363,7 @@ class AdminArea extends SFPComponent {
             <FilterHeader
               id='dtype'
               title={strings.countryTableDisasterType}
-              options={dTypeOptions}
+              options={[{ value: 'all', label: 'All Types' }, ...this.props.disasterTypesSelect ]}
```
GregoryHorvath, line 366: @frozenhelium @batpad I had to replace these in all places (~3-4) because I couldn't implement using the `disasterTypesSelect` from the `state` in the `utils/utils.js`

src/root/views/ThreeW/project-details.js
```diff
@@ -144,7 +144,10 @@ class ProjectDetails extends React.PureComponent {
               />
               <TextOutput
                 label={strings.threeWDisasterType}
-                value={dtype ? disasterTypes[dtype] : undefined}
+                value={dtype
+                  ? this.props.disasterTypesSelect.filter(dt => dt.value === dtype).label
```
GregoryHorvath, line 148: @frozenhelium @batpad Couldn't find out how to test this locally, I hope it works...

src/root/schemas/field-report-form.js
```diff
@@ -27,10 +27,10 @@ export const step1 = {
     districts: {
       type: 'array'
     },
-    disasterType: {
-      type: 'string',
-      enum: getValidValues(formData.disasterType, 'value')
-    },
+    // disasterType: {
```
GregoryHorvath, line 30: @frozenhelium @batpad Similar to `utils/utils.js`, I couldn't find a way to use the `disasterTypesSelect` array from the `state` here.

src/root/Multiplexer/index.js
```diff
@@ -193,13 +205,16 @@ const mapStateToProps = (state) => ({
   userResponse: userResponseSelector(state),
   languageResponse: languageResponseSelector(state),
   currentLanguage: currentLanguageSelector(state),
+  disasterTypes: state.disasterTypes,
+  disasterTypesSelect: disasterTypesSelectSelector(state),
```
GregoryHorvath, line 209: @frozenhelium @batpad I added `disasterTypesSelect` here, but simply whenever I tried to get it with `state.disasterTypesSelect` it was empty so thus why I just used `disasterTypesSelect: disasterTypesSelectSelector(state),` in all places.

src/root/utils/get-dtype-meta.js
batpad: Can we get the places that call `getDtypeMeta` to send in `disasterTypes` as a parameter to the function, so the function does not need to call the state directly, but just receives it as parameter - and I'm assuming the view or wherever `getDtypeMeta` is called from already has access to this from the state and can send it down as a param? Happy to look at this together - I agree that we should avoid accessing the store directly here if possible. Curious to see if @frozenhelium has a suggestion.

src/root/schemas/field-report-form.js
batpad: This is non-ideal, but okay, there should be no reasonable way for the user to get an invalid value here, and we are validating on the backend.

I would like to find a good pattern to get the schema validator to use values from the state, but this is probably a bit involved. If we figure that pattern out, we can implement it for countries as well, but I wouldn't worry too much about this, this seems fine.

src/root/views/ThreeW/project-details.js
batpad: @frozenhelium would you be able to help test this?

src/root/utils/get-dtype-meta.js
frozenhelium: Yep, I think we should pass in the `disasterTypes` as argument

src/root/Multiplexer/index.js
```diff
@@ -105,12 +109,20 @@ function Multiplexer(props) {
...
+    }
+  });
+
   const pending = React.useMemo(() => (
     (languageResponse.fetching || !languageResponse.fetched)
     || (allCountriesResponse.fetching || (!allCountriesResponse.cached && !allCountriesResponse.fetched))
     || (allRegionsResponse.fetching || (!allRegionsResponse.cached && !allRegionsResponse.fetched))
     || (!skipUserDetails && (userResponse.fetching || (!userResponse.cached && !userResponse.fetched)))
-  ), [allCountriesResponse, userResponse, allRegionsResponse, skipUserDetails, languageResponse]);
+    || (!disasterTypes || (!disasterTypes.cached && !disasterTypes.fetched))
```
frozenhelium, line 124: not sure but shouldn't it be 
```js 
|| (disasterTypes.fetching || (!disasterTypes.cached && !disasterTypes.fetched))
```

src/root/Multiplexer/index.js
frozenhelium: That's weird, the end result for both should be the same

src/root/Multiplexer/index.js
GregoryHorvath: @frozenhelium Somehow `fetching` and `fetched` were always `false` (except after I added `fetched: true` to the Multiplexer part).

src/root/views/ThreeW/project-details.js
GregoryHorvath: @batpad @frozenhelium I had to reproduce the same usage anyway and found out I need to use `.find()` here instead of `.filter()` and check if `.find()` is null-proof or not. Will be fixed soon.

src/root/Multiplexer/index.js
frozenhelium: That's okay, because we would not start the request if it is in the cache. So, `fetching` and `fetched` will mostly be `false` and `cached` will be true in that case

src/root/views/ThreeW/project-details.js
frozenhelium: @GregoryHorvath there might be new field `dtype_display`, we can directly use that instead (if there is one)

frozenhelium: It's actually `dtype_details?.name`

frozenhelium (approved):

src/root/utils/get-dtype-meta.js
GregoryHorvath: @frozenhelium @batpad I was able to completely remove the function alltogether.

src/root/views/ThreeW/project-details.js
GregoryHorvath: @frozenhelium thanks! Was able to use this at a couple of places.

GregoryHorvath: 44c22ff

src/root/schemas/field-report-form.js
GregoryHorvath: I'll ticket this out for later.

src/root/Multiplexer/index.js
GregoryHorvath: I'll ticket this out too for later

PR #1598, opened 2020-09-18
Fix/empty countryname

GregoryHorvath:
If the selected language doesn't have a translation for all the listed countries it gave an error `cannot read property [0] of null` at https://github.com/IFRCGo/go-frontend/blob/f93c1b1dcc3035827fa205166871aebe97b8b6df/src/root/components/country-list.js#L32


batpad:
@GregoryHorvath hmm I think it'd be better to not include the country name in the list if name is null - just something like https://github.com/IFRCGo/go-frontend/commit/d8f1eb47bddc07942a195ca22da710c6118ad29b - let me know if it makes sense to adapt this PR?

batpad (approved):

GregoryHorvath:
@batpad Thanks for the fix, included and merged :slightly_smiling_face:

Issue #1629, opened 2020-09-25
[PROD] Map 'flatness' and split countries that span the 180° date line


jhenshall:
## Issue
On all maps, the 180° date line acts as an 'edge' on either side meaning that maps cannot be centred on the Pacific. This is especially an visible when displaying the Asia Pacific region - e.g. Samoa cannot be included in the map view because it is on the far left of the map.

Also, Fiji and New Zealand (with outer islands) bounding boxes actually split the 180° line so there is weird behaviour of how these are stored in the backend and cannot be displayed in one view.

## Steps to reproduce
Go to any map and pan either east or west and you get to the 'edge'. e.g.:

![image](https://user-images.githubusercontent.com/6359748/94247132-9c9a2d00-ff14-11ea-838d-a19f84bfbb77.png)

## Expected behaviour
As you pan the map, it continues to wrap around the world and is not cut at 180 degrees. This will mean that maps can focus on the Pacific and you'd still see Asia to the west and Americas to the east. Ideally, bounding boxes could also span the date line (if possible).

## Related feature
- Initially reported here but split out as part of issue fixed - [Map centering on regional pages and map "flatness"](https://github.com/IFRCGo/go-frontend/issues/897)
- [Map centring on countries that span 180deg line](https://github.com/IFRCGo/go-frontend/issues/1162#issuecomment-697463562)

batpad:
cc @geohacker

batpad:
This was accidentally closed with commit message. Re-opening.

This is being deployed in v4.4.3.

PR #1630, opened 2020-09-25
Don't cut the map at 180

geohacker:
Fixes https://github.com/IFRCGo/go-frontend/issues/1629

cc @batpad  @jhenshall

### DIFF
--- a/src/root/utils/get-new-map.js
+++ b/src/root/utils/get-new-map.js
@@ -13,7 +13,7 @@ export default function newMap (container, mapStyle = 'mapbox://styles/go-ifrc/c
     scrollZoom: false,
     pitchWithRotate: false,
     dragRotate: false,
-    renderWorldCopies: false,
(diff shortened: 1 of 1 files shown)


geohacker:
@jhenshall if you'd like to test this https://ifrc-go-fix-map-at-180.surge.sh/

jhenshall:
Whoa @geohacker , so quick!! This is awesome. Seems good to go as a fix for the majority of the issue. Intrigued how regions and countries that split the 180 line can be handled in the backed?

geohacker:
>  Intrigued how regions and countries that split the 180 line can be handled in the backed?

Yeah this needs a bit more thinking but definitely doable :)

batpad (approved):

batpad:
Merging this as this fix definitely seems like an improvement? cc @geohacker @jhenshall

PR #1633, opened 2020-09-25
Checkie-checkie

GergiH:
Ref.: https://github.com/IFRCGo/go-frontend/issues/1548

Initially it looks like this:
image


batpad:
@frozenhelium trying to display an error message for IE users got a bit strange and am not sure if this is the best approach.

Could we add this to your backlog for the next sprint or so? I'm closing this PR as stale, but you should be able to check it out for reference / also just re-open it if you think this is a valid approach. Thanks.

PR #1650, opened 2020-10-13
Multilingual country label fixes

geohacker:
* [x] Only show country labels
* [x] Emergency detail view map labels


batpad:
Can be tested at https://ifrc-go-only-country-labels.surge.sh/

Issue #1675, opened 2020-11-18
[Prod] Banner - please remove the Survey banner 


tovari:
Please remove the GO Survey banner from the top of the site.
There is no request to replace it with another banner, it can be removed.

GregoryHorvath:
@tovari Please check around if everything works without the banner: https://ifrc-go-minor-remove-banner.surge.sh/

@batpad @geohacker Europe's page is def. not working for example, but not because of the banner removal (https://ifrc-go-minor-remove-banner.surge.sh/regions/3#operations). Most probably these get `Infinity` or `-Infinity` as a value as I've noticed one time.
```
Error: Invalid LngLat latitude value: must be between -90 and 90
    at new o (mapbox-gl.js:139)
    at Function.o.convert (mapbox-gl.js:139)
    at o.setSouthWest (mapbox-gl.js:141)
    at new o (mapbox-gl.js:141)
    at Function.o.convert (mapbox-gl.js:141)
    at t.fitBounds (mapbox-gl.js:452)
    at a.value (main-map.js:206)
    at a.value (map-component.js:74)
    at is (react-dom.production.min.js:212)
    at hu (react-dom.production.min.js:255)
```

batpad:
This is done and in production. Marking as Ready for UAT, please move to `Done` when confirmed.

batpad:
I am closing this as it's moved to Done and in production. @tovari please re-open if needed.

PR #1676, opened 2020-11-19
Comment out Global banner

GregoryHorvath:
Ref.: https://github.com/IFRCGo/go-frontend/issues/1675
@batpad this was branched from and points to `master`, not sure how urgent it is to merge, just wanted to make sure we can if we need to

# Changes
- Removes the Global banner at the top of the page

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+### Release 4.4.5 - Remove Survey Banner
+
+- Remove (hide) the Global banner: https://github.com/IFRCGo/go-frontend/issues/1675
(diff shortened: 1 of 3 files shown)


batpad:
@GregoryHorvath thanks!

Could you create a surge URL for this branch just to verify that nothing breaks with the layout or anything with the banner commented out?

batpad:
@GregoryHorvath could you also please increase the version number in package.json and make a one-line entry in the Changelog? Let me know if you want me to do that.

batpad (approved):

PR #1677, opened 2020-11-19
Iframe content for additional links

kamicut:
- [x] Adds click handlers for additional links
- [x] Iframes content according to click handlers
- [ ] looks for 'show_in_go' property
- [ ] styled iframe


batpad:
@kamicut this is looking good to me.

With this approach, I guess we don't store the state in the URL? It would be *nice* to have URL state for each of these "links", but if this adds a lot of complexity, I feel okay with proceeding like this and taking that up for discussion for later.

batpad:
This was incorporated as part of #1680 .

Seems to work well, thanks @kamicut !

PR #1683, opened 2020-11-25
fix display of emergency name in alerts table emergency link

batpad:
Fixes display of the Emergency names in the Alerts table for Alerts imported from Molnix (uses the name of the emergency if operation string is empty).


batpad:
cc @GregoryHorvath - going ahead and merging this as it's a pretty minor fix.

Issue #1202, opened 2020-05-06
Map - styling updated


mariam-yu:
Visual clean up of the map to pass on to devs, including review of the colored dots, legends, popup and such.

mariam-yu:
Design should tackle the following:

- Map - table correlation: Make the map and table correlation clearer that the filters apply to both and that visually they are connected.
- Maps are taking too much space: Reduce the size of the maps across the platform, may be besides home.
- Map table pagination vs View all: How much is shown in pagination and how much goes into view all?
- Map legends, filters and toggle
- Map colors and what they represent
- Map functionality - to be validated with dev.

mariam-yu:
Please take a look at some suggestions of the UI for the map. We are trying not to deviate much from the new branding guidelines, but it doesn't give much wiggle-room. 

// Option 1: proposing to remove the dots and use highlights on the map directly for Appeals, DREF, and mixed. We lose the scale element, but that was kind of missed anyway. The advantage is we can have comfortable hovers over every country to allow clicking through, even if there are no appeals/DREFs. 

Alternatively, we can use the same coloring to show the number of emergencies per country, while having Appeals and Drefs only in a dropdown.

![1 0_GO_Home  Map 1](https://user-images.githubusercontent.com/60006476/86244160-d4a90280-bba7-11ea-8857-899f7967996c.png)

// Option 2: improved color scheme based on new guidelines and cleaned up legends/dots
![1 0_GO_Home  Map 3](https://user-images.githubusercontent.com/60006476/86244173-da064d00-bba7-11ea-9dc9-9034fc100fe7.png)

// Option 3: lighter option to have the dots pop
![1 0_GO_Home  Map 4](https://user-images.githubusercontent.com/60006476/86244179-da9ee380-bba7-11ea-9c57-46b3861fba51.png)

// Option 4: playing with background vs back, making the background darker
![1 0_GO_Home  Map 2](https://user-images.githubusercontent.com/60006476/86244170-d8d52000-bba7-11ea-8b74-c1741d0035f7.png)

LukeCaley:
Thanks @mariam-yu.  Handing over to @jhenshall to move the MapBox styling along with @geohacker.  

- Personal preference is for concentric circles for lighter red for DREF, darker tone for Appeals and concentric circles of the two for Multiple Types.  This would give the sense of flow from small to large.  

[...]

PR #1702, opened 2020-12-03
#1202: Map popup styling for Home, Emergencies, Regions

imohkay:


batpad:
@geohacker could you please review and perhaps merge into your map styling branch to then be able to make a consolidated surge URL with the updated base map styles and popups? Thanks!

geohacker (approved):
Thank you @imohkay!

PR #1704, opened 2020-12-03
Map style updates

geohacker:
Fixes https://github.com/IFRCGo/go-frontend/issues/1202

### DIFF
--- a/src/root/components/map/home-map/explanation-bubble.js
+++ b/src/root/components/map/home-map/explanation-bubble.js
@@ -38,10 +38,12 @@ class ExplanationBubble extends React.Component {
               <Translate stringId='explanationBubbleType'/>
             </label>
             <dl className='legend__dl legend__dl--colors'>
+              {/* FIXME: this should be  #f63440 */}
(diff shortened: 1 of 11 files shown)


src/root/components/map/home-map/explanation-bubble.js
```diff
@@ -38,10 +38,12 @@ class ExplanationBubble extends React.Component {
               <Translate stringId='explanationBubbleType'/>
             </label>
             <dl className='legend__dl legend__dl--colors'>
+              {/* FIXME: this should be  #f63440 */}
```
geohacker, line 41: @imohkay @batpad what's the best way update these legend entries to match the circle colours on the map? Should I just add new css classes?

imohkay:
I think maybe add new css classes... There are colors in the variables partial... you can also the color to the variables first and then use in the legend  @geohacker

batpad:
@geohacker I fixed the popups for countries without emergencies - could you give a look, and let know if there's anything pending here, else let's merge to develop!

PR #1728, opened 2020-12-09
Fix missing months in chart

### LINKED ISSUE #1398: [PROD] Line graphs to show all months/years
## Issue
The EA/DREF and emergencies graphs on the home page / regional page / emergencies page show the number of these by month or year. In cases that there have been none on a year or month, instead of showing zeros, the graph just skips this leading to an inaccurate "timeline".

E.g. on the MENA refgional page (https://go.ifrc.org/regions/4#operations) in the graph below, April is missing

image

## Expected behaviour
All months or years should show on the timeline, even when there have been no operations or emergencies on that particular month/year

Sidenote that I suppose will be fixed with the new styling, but flagging for now: the "s" in DREFs should be lowercase
image

kamicut:
Closes #1398

### DIFF
--- a/src/root/components/timeline-charts.js
+++ b/src/root/components/timeline-charts.js
@@ -2,6 +2,7 @@ import React from 'react';
 import { PropTypes as T } from 'prop-types';
 import { connect } from 'react-redux';
 import { ResponsiveContainer, LineChart, Line, Legend, XAxis, YAxis, Tooltip } from 'recharts';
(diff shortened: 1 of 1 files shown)


batpad:
@kamicut - this looks good to me.

Could you remove yarn.lock from the PR though? I do think we have some issues to resolve with the yarn.lock, but maybe we can do that separately?

batpad (approved):

Issue #1740, opened 2020-12-11
Release 4.5.0 - mid-December, 2020


batpad:
Creating a ticket to try and bring all the small remaining action items for this release into a single place with clear assignees, and also to outline all the changes going into this release, to serve as the basis for the release notes and Changelog.

### Map Styling Updates: https://github.com/IFRCGo/go-frontend/issues/1202

Map styles and popups have been updated per design. Remaining actions:

 - ~~Remove the behaviour to highlight countries on hover (retain functionality to click on countries) - @geohacker~~
 - ~~Address @mariam-yu's feedback on Popup styling - https://github.com/IFRCGo/go-frontend/issues/1202#issuecomment-741780538 - @imohkay~~
 - ~~Adjust legend colours as per new colours on map / design (no need to adjust the legend design for this release, just make sure the colours are correct and correspond to the colours on the map) - @imohkay - to ask @geohacker for any clarifications.~~
 - Minor changes to map styles for disputed borders, etc. - @geohacker, is this actually pending?

### Geo-data Imports - https://github.com/IFRCGo/go-frontend/issues/1643

@geohacker and @jhenshall are working through CSVs and shape-files, mostly tracked here: https://github.com/IFRCGo/go-frontend/issues/1643#issuecomment-735389178

### Mapbox Vector Tile Update: https://github.com/IFRCGo/go-frontend/issues/1474

Backend work to generate tiles on Mapbox based on the spatial data in GO is complete and map tiles can be viewed on Mapbox. We will not make the frontend change to point production to the new tileset right away as it will need more validation, testing and approval. @jhenshall - it will be good if you could go through the generated tilesets on Mapbox and verify / confirm that the data is being output from GO correctly, and is complete.

### PER Refactor: https://github.com/IFRCGo/go-frontend/issues/1667

This has been a massive lift and @GregoryHorvath has led out completely re-writing the PER forms module, accommodating various changes to data structure and design based on feedback from the PER team. Basic functionality of filling in forms and saving, viewing and exporting has been tested. There are some remaining items TODO:

 - ~~Add subcomponents - @GregoryHorvath~~
 - ~~Add a permissions model so not all users can view all PER forms - @GregoryHorvath~~
 - ~~Fix styling for checkboxes - @GregoryHorvath to reach out to @imohkay with details of what is needed here.~~
 - ~~@GregoryHorvath - let me know anything else I'm missing and I'll add it here.~~
[...]

PR #1748, opened 2020-12-14
remove country on hover highlight on map, refs #1740

batpad:
cc @geohacker - I just commented out the map hover bits, could you review?

### DIFF
--- a/src/root/components/map/main-map.js
+++ b/src/root/components/map/main-map.js
@@ -233,9 +233,9 @@ class MainMap extends React.Component {
       }
     });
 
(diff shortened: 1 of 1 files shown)


geohacker (commented):
Could you comment this also https://github.com/IFRCGo/go-frontend/blob/2446a659b9c060a1a08d025bde2024a516d2435a/src/root/components/map/main-map.js#L236-L238 @batpad?

batpad:
@geohacker updated. can you check? thanks!

Issue #1749, opened 2020-12-14
[Staging] Missing headers across the site


batpad:
## Issue

When we added a `showHeader` option to the display-table, it broke headers for existing tables using the `DisplayTable` component.

@GregoryHorvath can we look at this together?

GregoryHorvath:
This was resolved with fixing the null-check of `showHeader`.

PR #1750, opened 2020-12-14
Fix nullchecking showHeader prop for DisplayTable

GregoryHorvath:
Ref.: https://github.com/IFRCGo/go-frontend/issues/1749

## Changes
- Fix nullchecking `showHeader` prop for `DisplayTable`, now defaults to `true`

### DIFF
--- a/src/root/components/display-table.js
+++ b/src/root/components/display-table.js
@@ -72,7 +72,7 @@ export default class DisplayTable extends React.Component {
   }
 
   render () {
-    const showHeader = this.props.showHeader ? this.props.showHeader : true;
+    const showHeader = this.props.hasOwnProperty('showHeader') ? this.props.showHeader : true;
     return (
       <React.Fragment>
         <table className={this.props.className}>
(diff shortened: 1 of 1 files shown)


batpad (approved):

src/root/components/display-table.js
```diff
@@ -72,7 +72,7 @@ export default class DisplayTable extends React.Component {
   }
 
   render () {
-    const showHeader = !!this.props.showHeader;
+    const showHeader = this.props.showHeader ? this.props.showHeader : true;
```
batpad, line 75: I'm not sure if that works since `this.props.showHeader` if false, would get set to true.

I think what you want is something like:

`const showHeader = this.props.hasOwnProperty('showHeader') ? this.props.showHeader: true;`

GregoryHorvath: Yea, sorry I just realized I totally skipped the `false` check with that. Pushing the fix in a sec.

batpad (approved):

PR #1756, opened 2020-12-15
make the labels black. make sure deployments map follow map style guide

geohacker:
Deployments map
image

Main map
image


batpad:
This is looking really nice!

batpad (approved):

PR #1773, opened 2020-12-16
poppins isn't rendering rtl well, so for arabic try to fallback to arial

geohacker:
@batpad please test.


batpad:
Works well for me!

PR #1783, opened 2020-12-16
Show no data message on Preparedness tab

GergiH:
cc @batpad


batpad (approved):

batpad:
This seems like something we would want -- I'm going to keep this PR open, let's merge it after this release and figure out a good way to test. cc @frozenhelium

GergiH:
@batpad This was most probably just a "hotfix" kind of change for the Preparedness tab for the Country pages, connected to the PER rework. But all in all we planned to rework this whole tab too, it just hadn't yet happened.

frozenhelium:
@batpad I've rebased it with latest develop, should we merge this?

batpad:
Thanks @frozenhelium

PR #1795, opened 2021-01-19
Epic Issues

GregoryHorvath:
This should add https://github.com/marketplace/actions/epic-issues-for-github to our repo (?)

@batpad 
cc @tovari

### DIFF
--- a/.github/workflows/update_epics.yml
+++ b/.github/workflows/update_epics.yml
@@ -1,0 +1,14 @@
+name: Update epics
+on:
+  issues:
+    types: [opened, created, closed, reopened, deleted]
+jobs:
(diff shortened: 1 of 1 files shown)


batpad:
@GregoryHorvath let's discuss this - the problem I foresee with this is the automatic marking of issues in the epic depend on the child issues is that we don't always close the issue from a developer stand-point when it's done (only after being tested on prod) - this might make these epic lists look a bit out of sync with reality. I like the idea, just not sure if it works perfectly with our current workflow.

batpad:
We are looking at other solutions to handle ticket dependencies, etc. Closing here.

Issue #1360, opened 2020-07-02
Make links to Admin in frontend point to new domain


batpad:
**Is your feature request related to a problem? Please describe.**

Currently, links to the Django Admin point to https://prddsgocdnapi.azureedge.net/ , which works, but it would be better to point to the more user-friendly https://goadmin.ifrc.org/

Currently, we use the same config variable called `api` to construct URLs for API calls as well as the Admin.

**Describe the solution you'd like**

Options would be:

 - Create a new config param like `adminUrl` and have it default to the API url.
 - Move over the API end-points also to https://goadmin.ifrc.org/ as the domain - while this would work, I'd rather not change that.

So, we'll likely just have to create a new config parameter and use that.

xref #1116

batpad:
Will be good to do as part of general cleanup. Moving to 4.4.

tovari:
Has this been completely resolved, @batpad?

PR #1803, opened 2021-01-25
change links to admin pages to use adminUrl

kamicut:
Changes `href` links to use adminUrl. I’m not sure about what the default should be for non production links so I added the staging site, is that correct @batpad?

### DIFF
--- a/src/root/config.js
+++ b/src/root/config.js
@@ -1,6 +1,7 @@
 export const environment = process.env.REACT_APP_ENVIRONMENT || process.env.NODE_ENV || 'development';
 export const api = process.env.REACT_APP_API_ENDPOINT ||
   (process.env.NODE_ENV === 'production' ? 'https://prddsgocdnapi.azureedge.net/' : 'https://dsgocdnapi.azureedge.net/');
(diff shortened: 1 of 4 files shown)


batpad (approved):

batpad:
This refs https://github.com/IFRCGo/go-frontend/issues/1360 - part of moving URLs in the frontend away from the azureedge.net URL.

@GregoryHorvath could you give this a look and merge, else I'll go ahead and merge this soon.

src/root/config.js
```diff
@@ -1,6 +1,7 @@
 export const environment = process.env.REACT_APP_ENVIRONMENT || process.env.NODE_ENV || 'development';
 export const api = process.env.REACT_APP_API_ENDPOINT ||
```
GregoryHorvath, line 2: We probably needed this check because this is being used for local development I believe, but I must be wrong. @batpad Sorry don't have the time to test right now but please make sure we're not messing with prod data locally even by accident.

batpad: @kamicut could you look? happy to discuss.

batpad: @GregoryHorvath this looks okay to me, can we deploy this to staging and test? This doesn't seem to change the fundamental logic of things apart from changing the URL.

GregoryHorvath: @batpad Sure we can merge, but I'd say still keep an eye on local development. If that's untouched we're good, but I highly suspect because of this change also the local development will point to the staging DB, but again I didn't have the time to test this.

PR #1847, opened 2021-03-02
revise map marker colours and match legend

geohacker:
Address https://github.com/IFRCGo/go-frontend/issues/1202#issuecomment-785016473

image

cc @jhenshall @batpad @GregoryHorvath


jhenshall (approved):
Nice one! (Could have tried that one myself) 🙂

PR #1903, opened 2021-05-06
feat: setup application insights go-api#1018

gulfaraz:
This PR sets up rudimentary analytics tracking on the go-frontend using [MS Application Insights](https://docs.microsoft.com/en-us/azure/azure-monitor/app/app-insights-overview).

Please add the new environment variable to all go-frontend deployments.
`REACT_APP_INSTRUMENTATION_KEY=fa128365-fb09-4568-b944-b05e38b6b94c`

Please see https://github.com/IFRCGo/go-api/issues/1018 for details.


batpad:
@gulfaraz really sorry I missed this.

Would we all need to add this as env var locally as well? Or should we use a different / dummy key or so?

@frozenhelium sorry - just a heads-up about this PR so we can figure with @gulfaraz on dealing with eventual merge conflicts. Should all be fine.

gulfaraz:
@batpad A local variable for this may not be needed. If it throws an error, then any non-empty value would simply lead to 400 errors for the analytics logging.

batpad:
@gulfaraz does the decision to use G analytics on the frontend https://github.com/IFRCGo/go-api/issues/1018#issuecomment-845134993 affect this PR / is this still needed?

gulfaraz:
@batpad Yes, it definitely does. You beat me to the comment by 43 seconds!

PR #1945, opened 2021-06-18
Feature/global 3w charts

frozenhelium:


src/root/views/GlobalThreeW/Charts/index.tsx
```diff
@@ -0,0 +1,174 @@
...
+  key: number,
+  name: string,
+  value: number,
+}
+
+const PIE_COLORS = ['#f64752', '#fa999f', '#f87079'];
+
+
+interface PieChartProps {
+  data: any[],
```
tnagorra, line 29: We shouldn't use `any` type

src/root/views/GlobalThreeW/Charts/index.tsx
```diff
@@ -0,0 +1,174 @@
...
+        return {
+          key: datum["programme_type"],
+          name: datum["programme_type_display"],
+          value: datum["count"],
+        };
+      });
+      return dataArr;
+    }
+
+    return [];
```
tnagorra, line 51: It's better to shortcircuit the if statement

src/root/views/GlobalThreeW/Charts/index.tsx
```diff
@@ -0,0 +1,174 @@
...
+        className={styles.chartContainer}
+      >
+        <PieChart>
+          <Pie
+            data={chartData}
+            dataKey="value"
+          >
+            { data.map((entry, index) => (
+              <Cell
+                key={`cell-${index}`}
```
tnagorra, line 69: It's not a good idea using index as key

src/root/views/GlobalThreeW/Charts/index.tsx
```diff
@@ -0,0 +1,174 @@
...
+        className={styles.tick}
+      >
+        {value}
+      </text>
+    </g>
+  );
+}
+
+interface BarChartProps {
+  data: any[],
```
tnagorra, line 108: We shouldn't use `any` type

src/root/views/GlobalThreeW/Charts/index.tsx
```diff
@@ -0,0 +1,174 @@
...
+      const dataArr = data.map((datum: any) => {
+        return { key: datum[keys[0]], name: datum[keys[1]], value: datum["count"] };
+      });
+
+      return dataArr
+        .sort((a, b) => b.value - a.value)
+        .splice(0, 5); // Only show top 5
+    }
+
+    return [];
```
tnagorra, line 131: Better to shortcircuit the if statement

src/root/views/GlobalThreeW/Charts/index.tsx
```diff
@@ -0,0 +1,174 @@
...
+function GoBarChart(props: BarChartProps) {
+  const {
+    data,
+    className,
+  } = props;
+
+  const chartData = React.useMemo(() => {
+    if (data && data.length > 0) {
+      const keys = Object.keys(data[0]);
+      const dataArr = data.map((datum: any) => {
```
tnagorra, line 122: The type shouldn't be `any`

src/root/views/GlobalThreeW/Charts/index.tsx
```diff
@@ -0,0 +1,174 @@
...
+        return { key: datum[keys[0]], name: datum[keys[1]], value: datum["count"] };
+      });
+
+      return dataArr
+        .sort((a, b) => b.value - a.value)
+        .splice(0, 5); // Only show top 5
```
tnagorra, line 128: Shouldn't this be configurable?

tnagorra (changes requested):

barshathakuri (approved):

src/root/views/GlobalThreeW/Charts/index.tsx
frozenhelium: This is only for the Global 3W page, so I guess it's okay

Issue #1837, opened 2021-02-22
[PROD] Today line in deployments Surge tab in an emergency not showing correctly


anamariaescobar:
## Issue
The today line in the section of the rapid response personnel deployed in the surge tab of an emergency is not showing correctly.

![image](https://user-images.githubusercontent.com/69478743/108722654-ad7eb700-7523-11eb-9ce6-3ad2c0f3dfc2.png)

## Steps to reproduce
https://go.ifrc.org/emergencies/5022#surge

anamariaescobar:
Update:

The today line is still behaving crazy:

![image](https://user-images.githubusercontent.com/69478743/123412112-84456480-d5b1-11eb-838d-8014cb54e2ec.png)

The today line should be aligned to the dates in the bottom, this makes it look like the deployments were completed when actually they are on going, it would be great to have the today line more to the left (still matching the dates in the bottom) so we can see to the future of the operations if there will be any gaps in personnel.

Link to one example of the issue: https://go.ifrc.org/emergencies/5308#surge

@tovari

batpad:
@anamariaescobar is it by any chance possible to get a link on staging where this can be observed?

Thanks much.

batpad:
My hunch is this might be a bug when the results are paginated / some improper handling of the "today" placement calculation when all the results are not on a single page / we might not have tested this case properly. If there is an example I can reproduce on staging, I'll prioritize looking into this.

anamariaescobar:
I found one @batpad ! https://dsgoproxyapp.azurewebsites.net/emergencies/202#surge and it doesn't seem to have any pagination.....

anamariaescobar:
hey @batpad just found another issue, not sure if it is related, there is no way to click on the pagination for deployments, for example in this page:
https://go.ifrc.org/emergencies/5308#surge

PR #1954, opened 2021-07-01
fix positioning of today line on surge table, refs #1837

batpad:
Fixes the positioning of the line representing "today" on the Personnel Deployments table / chart, refs #1837. This is a visual / CSS fix and should be safe to merge in.

@frozenhelium - could you give a review? Knowing that the overall code for this complex chart squeezed into the existing table structure is all a bit terrible - would be nice to have your eyes on it for a future refactor / to keep in mind whenever we are refactoring the tables and filters - for now, this just fixes the bug of the line appearing in the wrong place, so hopefully is fine to merge. Thanks!


frozenhelium (approved):
Code looks good, haven't run it though!

PR #1955, opened 2021-07-02
Fix 3W feedbacks

frozenhelium:


src/root/views/GlobalThreeW/index.tsx
```diff
@@ -251,15 +257,22 @@ function GlobalThreeW(props: Props) {
...
-              to="/regions/0#3w"
-              {...exploreRegional3WLinkProps}
-            />
+            <DropdownMenu
+              label={<div {...exploreRegional3WLinkProps} /> }
+            >
+              {regionList.map((region) => (
+                <DropdownMenuItem
+                  label={region.region_name}
+                  href={`/regions/${region.id}#3w`}
```
tnagorra, line 272: `key` prop is missing

src/root/views/GlobalThreeW/index.tsx
```diff
@@ -69,6 +72,9 @@ function GlobalThreeW(props: Props) {
     location,
   } = props;
 
+  const allRegions = useReduxState('allRegions');
+  const regionList = allRegions?.data?.results || [];
```
tnagorra, line 76: Lets not create a new object on render function.
Why not just use optional chaining on regionList instead

AdityaKhatri (approved):

PR #1957, opened 2021-07-05
Fix minor issues in 3W

frozenhelium:


src/root/views/Country/ThreeW/InCountryProjects/index.tsx
```diff
@@ -43,19 +43,19 @@ import {
 import ProjectFormModal from '../ProjectFormModal';
 import ProjectTableActions from '../ProjectTableActions';
 import ProjectStatPieChart from '../ProjectStatPieChart';
-import ProjectFlowSankey from '../ProjectFlowSankey';
+// import ProjectFlowSankey from '../ProjectFlowSankey';
```
AdityaKhatri, line 46: Why remove Sankey?

AdityaKhatri (approved):

src/root/views/Country/ThreeW/InCountryProjects/index.tsx
frozenhelium: It was one of requirement

PR #2000, opened 2021-08-16
Use default empty array for molinx-tags if not provided.

thenav56:
Address surge page not working because of missing molinx-tags from API.

NOTE: A hotfix has already been deployed to production (master). https://github.com/IFRCGo/go-frontend/pull/2001

Error page: https://go.ifrc.org/emergencies/5499#surge


batpad:
@thenav56 thank you so much for this, and apologies for not catching this.

I have merged the hotfix made to master back into develop, so this change now exists in develop. Am closing this PR. Thanks again for the detailed explanation above.

Issue #1965, opened 2021-07-07
Frontend Changes to Surge pages for Molnix Integration


anamariaescobar:
Changes in the frontend, specifically Global surge page: [https://go.ifrc.org/deployments](url) and the Emergency surge page (example): [https://go.ifrc.org/emergencies/5308#surge](url) needed for the soon deployment of the Molnix system (Deployment date 1 of August 2021).

Wireframes prepared by Mariam here: [https://projects.invisionapp.com/share/B3XKVKZJWY9#/screens/361601984_5-1_GO_Deployment](url) and [https://projects.invisionapp.com/share/3QXKWXWTAHF#/screens/414729370_4-3-4_GO_Specific-Emergency_Epi](url)

Some point to take into account:

**Global Surge page**
- [x] #2011
- [x] Add a new key figure with Total Rapid Response personnel deployed (this will be a cumulative number of all deployments for the year)
- [ ] On the little information button on the key figures we should add a pop up that reads: Currently deployed, for both Rapid Response personnel and ERU.
- [x] Map: I think we should follow the style of the map on the front page of GO [https://go.ifrc.org/](url), the scale of the points will be the number deployed, categories will be Rapid Response Personnel and ERU.
- [x] Charts below the map: Top 5 NS deploying Rapid Response Personnel, the tittle should include the year and the chart must be cumulative for the present year
- [x] Charts below the map: Rapid Response personnel, add in the last 12 months in the tittle and the chart should always show the last 12 months of data.
- [x] Open Surge alerts: The field application deadline can be ignored, the surge team has decided not to input any data there.
- [x] Open Surge alerts: Can we move the column Start Date before Duration?
- [x] Keywords: Is there a way to filter specific tags from showing in GO (example the ops tags, there is a couple specific ones that they are not interested in showing here)
- [x] Deployments Overview by Emergencies: Can we change this tittle to: Active Deployments Overview by Emergencies?

**Emergency Surge Page**
- [x] Can we add a new Key figure "Deployed ERU"? This way all the key figures will match between global and emergency surge pages.
- [x] Can we make sure the style of the Key figures is the same for both  global and emergency surge pages?
- [x] Open Surge alerts: The field application deadline can be ignored, the surge team has decided not to input any data there.
- [x] Open Surge alerts: Can we move the column Start Date before Duration?
[...]

PR #2012, opened 2021-08-26
Update Surge page map with ERUs and Personnel data

willemarcel:
Contributes to #1965

### DIFF
--- a/src/root/components/map/deployments-map.js
+++ b/src/root/components/map/deployments-map.js
@@ -1,180 +1,117 @@
 import React from 'react';
-import * as path from 'path';
 import { render } from 'react-dom';
 import { PropTypes as T } from 'prop-types';
 import mapboxgl from 'mapbox-gl';
-import chroma from 'chroma-js';
-import _cloneDeep from 'lodash.clonedeep';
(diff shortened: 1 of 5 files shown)


willemarcel:
> Hi @willemarcel, @batpad. Looks good, but I think, the colors need to be aligned with those on the active operations map:
> #ff9e00
> #4c5d9b
> Thanks!

@tovari I pushed an update with the new map colours and deployed it: https://upbeat-bartik-c5abb7.netlify.app/deployments

tovari:
Hi @willemarcel, I think the displayed ERUs on the map are filtered by the ERU list pages, i.e. only those are visible that are on the actual page of the list. I could not check this for deployments as there is only 1 page at the moment. 
Check USA on the screenshots.
Further issue is that there is no mixed item is displayed on the map (USA should be).
Also, when clicking on page 3, the page breaks.
surge-map-1b

surge-map-2

willemarcel:
Thanks for reporting the bugs, @tovari. I have already fixed them. You can check on the same URL.

batpad:
Thanks @willemarcel - am going to go ahead and merge this to develop - we can make any additional fixes as separate branches.

PR #2014, opened 2021-08-27
Add form validation

barshathakuri:
Add validation in form


src/root/views/DrefApplicationForm/DrefOverview/index.tsx
```diff
@@ -210,7 +200,7 @@ function DrefOverview(props: Props) {
             options={yesNoOptions}
             radioKeySelector={booleanOptionKeySelector}
             radioLabelSelector={optionLabelSelector}
-            value={value.emergency_appeal_planned}
+            value={value.type_of_onset == 3 ? value.emergency_appeal_planned = false : value.emergency_appeal_planned}
```
frozenhelium, line 203: We should define a constant instead of directly comparing with value

`const ONSET_IMMITENT = 3`
```suggestion
            value={value.type_of_onset == ONSET_IMMITENT ? value.emergency_appeal_planned = false : value.emergency_appeal_planned}
```

PR #2018, opened 2021-08-30
Add "Deployments over the last year" chart to the Surge page

willemarcel:
- Create separate component TimeLineChart and use it both on deployments and in emergencies page

PR deployed at https://ifrc-go-new-deploy-chart.surge.sh

### DIFF
--- a/src/root/components/Charts/index.js
+++ b/src/root/components/Charts/index.js
@@ -1,0 +1,45 @@
+import React from 'react';
+
+import { ResponsiveContainer, LineChart, Line, XAxis, YAxis, Tooltip } from 'recharts';
+import { DateTime } from 'luxon';
+import Translate from '#components/Translate';
+
+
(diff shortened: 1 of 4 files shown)


frozenhelium:
@willemarcel We already have `useRequest` hook for fetching data. 
Example: https://github.com/IFRCGo/go-frontend/blob/32407f71c95318bb998cf30523ea4da9af88006b/src/root/components/ThreeWForm/useThreeWOptions.ts#L174-L180
CC: @batpad

batpad:
Thanks so much for the catch and review @frozenhelium -

@willemarcel does this make sense and would you be able to make this change?

willemarcel:
@batpad @frozenhelium I've made the change in order to use `useRequest`.

batpad (approved):

batpad:
This looks good to me and works well.

@frozenhelium could you take a final look? Thanks.

Issue #2035, opened 2021-09-06
Remove Type filter from Deployments Table


batpad:
Refs #1965 

We should remove the `Type` filter on the Deployments table, as we no longer have multiple deployment types.

cc @szabozoltan69

szabozoltan69:
https://github.com/IFRCGo/go-frontend/pull/2036

PR #2036, opened 2021-09-06
Not allowing type filtering in RRP table

szabozoltan69:
Solves https://github.com/IFRCGo/go-frontend/issues/1965#issuecomment-912888790 "white page" issue and https://github.com/IFRCGo/go-frontend/issues/2035


szabozoltan69:
I've put to comment all the not used translations.

batpad:
@szabozoltan69 we will also need to remove the Type arrow dropdown from the header here: https://github.com/IFRCGo/go-frontend/blob/develop/src/root/components/connected/personnel-table.js#L184 (just make it a simple string and not a FilterHeader)

And also we need to do the same thing for the `all personnel` table, like here: https://github.com/IFRCGo/go-frontend/blob/develop/src/root/components/connected/personnel-table.js#L184

batpad:
@szabozoltan69 could you also implement the same on the similar, but different `all-personnels` table: https://github.com/IFRCGo/go-frontend/blob/develop/src/root/components/connected/all-personnel-table.js#L145 (Sorry, I see I had the wrong link in my comment above!)

szabozoltan69:
Done for all personnel also.
kép

batpad (approved):

batpad:
Looks good to me, thanks @szabozoltan69 - once it's on staging, can ask folks to test and verify.

szabozoltan69:
@batpad – It's on staging.
kép

szabozoltan69:
kép

PR #2037, opened 2021-09-06
Commenting not used frontend translations

szabozoltan69:
I created it programmatically (not  manually searching). Tested thoroughly.


szabozoltan69:
Maybe we could put the not used ones into an other file... just for the case of "again needed".

PR #2047, opened 2021-09-08
Hotfix for ERUs not showing on the map

batpad:
Hotfix for ERUs not displaying on the map.

The logic for fetching the different ERU information does seem a bit convoluted and we should inspect it better.

This fixes the display on the map though.

cc @szabozoltan69


szabozoltan69 (approved):
Great fix

Issue #2042, opened 2021-09-07
[PROD] ifrc.org upgrade means Country weblinks no longer work and some other links on GO are broken too


nanometrenat:
## Issue
Weblinks to IFRC.org country pages were added via #776, and further enhanced via #922.  Every country page has a footer that shows links to that country's page on ifrc.org.

Or at least, it did until the ifrc.org website recently got upgraded - meaning that the old links no longer work.

e.g. DPRK page on GO https://go.ifrc.org/countries/193 - the `Democratic People's Republic of Korea on IFRC.org` button links to `https://www.ifrc.org/en/what-we-do/where-we-work/asia-pacific/red-cross-society-of-the-democratic-peoples-republic-of-korea/` which just gives `Page not Found`
![image](https://user-images.githubusercontent.com/33625192/132275804-cd941e83-31f0-4c50-aa76-659991a2c54b.png)

Similar for other countries e.g. https://go.ifrc.org/countries/165 for Switzerland sends you to broken https://www.ifrc.org/en/what-we-do/where-we-work/europe/swiss-red-cross/

## Steps to reproduce
Go to any country page and click on the relevant button in bottom left

## Expected behaviour
Links should work

## Related feature
#776

## Impact
All users/all countries

## Severity
It's annoying, and makes the button pointless - but the workaround is to search on the IFRC website.

## What's required
IM will need to manually update the website links in the back end for each country.  OR figure out what the new ifrc.org URL structure is and make the links follow that logic.

nanometrenat:
I've also seen that there are other places on GO where links to the ifrc.org website don't work now - e.g. the 'for more information on the PER Approach select **here**.' link on https://go.ifrc.org/preparedness#global-summary, which directs to `https://media.ifrc.org/ifrc/what-we-do-disaster-and-crisis-national-society-preparedness-effective-response/` (now broken).

Also at least one link on the About page.  Not sure about others - someone will need to go through and check all the ifrc.org weblinks and make sure they work (some will be able to be amended in the back end by a business user, others will presumably need a code change).

All the best

szabozoltan69:
The NS pages ( = country's page on ifrc.org) has this structure on ifrc.org:
https://www.ifrc.org/national-societies-directory/swiss-red-cross
https://www.ifrc.org/national-societies-directory/german-red-cross
and so on. I begin to fix them.

tovari:
The Reference centres link on the Resource page should point to
https://www.ifrc.org/reference-centres

[...]

PR #2052, opened 2021-09-14
Fixing broken links

szabozoltan69:
The country page related links should be fixed via database (via some api_country.url_ifrc updates).


szabozoltan69:
Solves https://github.com/IFRCGo/go-frontend/issues/2042

PR #2054, opened 2021-09-15
Feature/surge tab timeline catalogue

tamasmajorAtos:
- New tab under the Surge page
- Catalogue of Services
- Operational Toolbox


szabozoltan69 (approved):
Great job.

batpad:
@tamasmajorAtos really nice work!

There's a lot to go through and there's likely some parts we can think of cleaning up as we go along - for eg. it might be nice to think of a good way to break up the `data` variables into separate files.

Also I wonder if we can make some of the event handlers on the SVG a bit easier to understand - but I think we can do those down the line - this looks really solid to me and works really well - this kind of thing was always going to be a bit complex to implement, and this looks good to me. In the future, if we are moving a lot of this configuration to the backend, maybe the `data` can come from API calls.

This looks like a great start to this. Again, really nice work.

tamasmajorAtos:
@szabozoltan69 @batpad Thank you, for the kind words.

I updated code as you suggested. I moved the data variables to a separate files and clean up some code parts to be more understandable.

tamasmajorAtos:
> @szabozoltan69 @tamasmajorAtos - I see a couple files got checked in that maybe should not be part of this PR?
> 
> .vscode/launch.json and .vs/slnx.sqlite
> 
> Could we remove these?

Yes, I agree we should remove those files. Or extending the .gitignore file for ignore the .vscode and .vs folders

batpad:
> Yes, I agree we should remove those files. Or extending the .gitignore file for ignore the .vscode and .vs folders

Adding to `.gitignore` sounds perfect. Would you be able to make another quick PR with this @tamasmajorAtos ? Thanks so much!

tamasmajorAtos:
new PR for gitignore #2081

PR #2058, opened 2021-09-18
Eliminate linter and not-used warnings

szabozoltan69:


batpad:
@szabozoltan69 this looks good :+1: - I think we can just remove the code instead of commenting it out.

szabozoltan69:
Ok, I remove the unused rows.

PR #2081, opened 2021-10-13
remove .vs and .vscode folders

tamasmajorAtos:
@batpad 
Update .gitignore file for ignoring the .vs and .vscode folders.


batpad:
Thanks @tamasmajorAtos !

PR #2158, opened 2021-12-07
Update terser webpack plugin

frozenhelium:


.circleci/config.yml
```diff
@@ -5,11 +5,16 @@ jobs:
     docker:
       - image: node:15
 
+    when:
+      condition:
+        or:
+          - equal: ["${CIRCLE_BRANCH}", "${STAGING_BRANCH}"]
+          - equal: ["${CIRCLE_BRANCH}", "${PRODUCTION_BRANCH}"]
+      resource_class: large
```
szabozoltan69, line 13: I'm not sure that large resource class is available in our tier.

frozenhelium: It is apparently, I was surprised as well.

PR #2159, opened 2021-12-09
Move tabs to using React Routes

batpad:
Very WIP - change tabs to use React routes to manage routing, so that we can handle nested URL state when there is navigation within tabs.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2
 jobs:
   build:
     docker:
-      - image: node:14
+      - image: node:15
 
     resource_class: large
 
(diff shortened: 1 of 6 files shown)


batpad:
@szabozoltan69 i think this is close to ready - could you do a test:

 - Tabs on /deployments should show correctly and work
 - Current tab should be highlighted correctly
 - Should go to /deployments/overview by default
 - Sub-navigation inside Catalogue of Services should work

@frozenhelium would be good to get a code review from you, based on our discussion -- am happy to potentially convert this to tsx or so - let me know what sounds good / if the general idea / pattern here seems okay or if you would do this better / differently. We can maybe make some of the improvements later, but would be good to have your thoughts.

PR #2175, opened 2022-01-03
Feature/informal update

puranban:


src/root/views/InformalUpdateApplicationForm/Report/styles.module.scss
```diff
@@ -0,0 +1,66 @@
...
+.situational {
+  :global {
+    .go-container {
+      header {
+        background-color: #fff;
+        padding: $spacing $spacing-2 !important;
+      }
+      .styles_content__3hMfk {
+        padding: 0px;
+      }
```
frozenhelium, line 12: This classname will change on compilation, so wouldn't work

src/root/views/InformalUpdateApplicationForm/Report/styles.module.scss
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  &-actionTaken {
+    :global {
+      .styles_content__3hMfk{
+        background-color: $white;
+        padding: $spacing-2 !important;
+      }
+    }
+  }
```
frozenhelium, line 37: We should not style like this

frozenhelium (commented):
Also, only use 2 spaces for formatting

Issue #2172, opened 2021-12-23
[PROD] Map export not working


jhenshall:
## Issue
On emergency page maps (e.g. [this page](https://go.ifrc.org/emergencies/5140)), when clicking the export button it first exports a PDF with only the logos and legend, but with no map. Press the button again and it exports the map only, but with no logos or legend. This is obviously a bug and should be a combination of the map with the logos and legend every time button clicked.

## Steps to replicate

Click the button once and this is exported:
![2021-12-15-undefined-Brazil](https://user-images.githubusercontent.com/6359748/147232135-db969e39-d95b-4652-b6b0-ce286c5cfb6b.png)

Click the same button again and this is exported:
![2021-12-15-undefined-Brazil (1)](https://user-images.githubusercontent.com/6359748/147232280-f0a5c16a-b1b6-4ef0-9bfb-344ea99cf3a3.png)

## Related tickets
Not related to this bug, but if working on map export, might be worth also fixing the undefined file names: #1787

szabozoltan69:
It's true that the error exists on production and even on staging (e.g. https://dsgoproxyapp.azurewebsites.net/emergencies/4377 ), but locally I could not reproduce it: https://porgeto.hu/tmp/map_export1.webm
So it can be a consequence of some external lib, not the code itself.
Anyway, that "undefined" in filename should be fixed, so I do it now.

batpad:
I can reliably reproduce this across browsers - this is quite mysterious.

If unable to reproduce locally, this makes it a bit trickier.

@szabozoltan69 let me know if you have any luck - I'll try and think about this a bit and take a look at the code and see if I can identify what might be causing this - but this is really weird behaviour and I don't have any great ideas right away.

batpad:
cc @geohacker just in case you have any ideas

szabozoltan69:
The "undefined" filename part is fixed on staging environment, but the strange behavior still exists:
https://github.com/IFRCGo/go-frontend/pull/2181
(Try it via https://dsgoproxyapp.azurewebsites.net/emergencies/4377 )

szabozoltan69:
Now works fine on staging, e.g. https://dsgoproxyapp.azurewebsites.net/emergencies/4377

PR #2182, opened 2022-01-07
add setTimeout to try and fix map export issue

batpad:
@szabozoltan69 could you try and see if this fixes the issue for you as I am unable to reproduce locally

refs #2172


szabozoltan69 (approved):
Works fine in local tests.
(A bit long is the "flash up" of the big logo, but no problem. Better, than the non-working version. And we could create a just-fine-size logo which is not so big.)

Issue #2179, opened 2022-01-05
[PROD] Surge Emergency page country deployed to wrong


anamariaescobar:
## Issue
In the Rapid Response Personnel section in the Surge tab of an Emergency page, the Country deployed to is showing the country linked to the Emergency page, which is not always correct when we have multiple countries emergencies. We had solve this issue by displaying the country deployed to that was being pulled from Molnix, perhaps the las released set it back to before the fix.

![image](https://user-images.githubusercontent.com/69478743/148238199-464d05c3-ef74-4341-a38d-879c2f82485d.png)

## Steps to reproduce
Go to this Url https://go.ifrc.org/emergencies/5567#surge and hover the gant bar, this person should be deployed to Tajikistan and not Switzerland.

## Expected behaviour
The country the person is deployed to should be pulled form Molnix and not from the one associated with the Emergency page.

## Impact
All users trying to access the deployment information from operations with multiple countries.

## Severity
Functional impact to a field operation

## Is there a workaround?
There is no manual walk around this issue.

@tovari @batpad @szabozoltan69

batpad:
> We had solve this issue by displaying the country deployed to that was being pulled from Molnix, perhaps the las released set it back to before the fix.

@anamariaescobar are you sure this was ever fixed? As far as I know, this was discussed, but we have not implemented this yet. Really sorry, this likely slipped through the cracks.

Right now the logic is still that Deployments are associated with emergencies, and the country is pulled from the country of the emergency.

The implementation could maybe use a bit of discussion, but I think there's roughly two options here:

### Option One: New Column

 - Add a new column to the PersonnelDeployment model like `molnix_countries` or so: https://github.com/IFRCGo/go-api/blob/develop/deployments/models.py#L137

### Option Two: Migrate country -> countries

 - The second option would be to change the existing `country_deployed_to` column on PersonnelDeployment to be a ManyToMany relation instead of a ForeignKey - so make it like `countries_deployed_to` and migrate the existing data, and all usage of this through the app to expect a list instead of a single country.

Option Two does seem better, but I'm a bit worried about how much we may need to change if we allow Deployments to be deployed to multiple countries - Option One does seem like a bit of a hack, but is likely easier to implement.

For either of the two options, we'd need to:

 - Modify the `sync_molnix` management command to read countries from the Molnix API, instead of reading from the emergency the deployment is associated with: https://github.com/IFRCGo/go-api/blob/develop/api/management/commands/sync_molnix.py#L148
[...]

PR #2198, opened 2022-01-27
show country_to from individual personnel table, refs #2179

batpad:
Shows the Country Deployed to from the Personnel table, that comes from Molnix.

If that data does not exist, it falls back to showing the country of the emergency the deployment is associated with, so should continue working for old data.

cc @szabozoltan69 there do not seem to be any deployments in staging so was having a hard time testing this exactly. We may need to create some deployments in Molnix and simulate on staging to test that this works as intended - cc @anamariaescobar .

### DIFF
--- a/src/root/components/connected/personnel-table.js
+++ b/src/root/components/connected/personnel-table.js
@@ -240,7 +240,7 @@ class PersonnelTable extends SFPComponent {
                         </div>
                         <div class="row-sm spacing-half-b flex">
                           <div class="col-sm">Deployed To</div>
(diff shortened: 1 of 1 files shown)


batpad:
@szabozoltan69 I have not been able to test this thoroughly locally. But I think I feel okay about merging this and then asking / coordinating with @anamariaescobar to perform a full test with the Molnix data etc from Molnix staging. If this looks okay to you, could you merge?

szabozoltan69:
I created usable test data (with living deployments) on staging.

anamariaescobar:
Anything I can do to help with the testing? @szabozoltan69 @batpad

PR #2202, opened 2022-01-28
Using country_to in all-personnel-table also

szabozoltan69:


src/root/components/connected/all-personnel-table.js
```diff
@@ -151,7 +151,7 @@ class AllPersonnelTable extends SFPComponent {
         role: get(o, 'role', nope),
         type: o.type === 'rr' ? typeLongNames[o.type] : o.type.toUpperCase(),
         country: o.country_from ? <Link to={`/countries/${o.country_from.id}`} className='link--table' title={strings.personnelTableViewCountry}>{o.country_from.society_name || o.country_from.name}</Link> : nope,
-        deployed: o.deployment && o.deployment.country_deployed_to ? <Link to={`/countries/${o.deployment.country_deployed_to.id}`} className='link--table' title={strings.personnelTableViewCountry}>{o.deployment.country_deployed_to.name}</Link> : nope,
+        deployed: o.country_to ? <Link to={`/countries/${o.country_to.id}`} className='link--table' title={strings.personnelTableViewCountry}>{o.country_to.name}</Link> : (o.deployment && o.deployment.country_deployed_to ? <Link to={`/countries/${o.deployment.country_deployed_to.id}`} className='link--table' title={strings.personnelTableViewCountry}>{o.deployment.country_deployed_to.name}</Link> : nope),
```
batpad, line 154: @szabozoltan69 can you refer to the way I did it in the other PR? We should not just call `o.country_to.name` in case the country_to does not exist. It should use the country from the deployment in case `o.country_to` is undefined.

batpad: @szabozoltan69 like this: https://github.com/IFRCGo/go-frontend/blob/develop/src/root/components/connected/personnel-table.js#L243

szabozoltan69: I copied the solution from yours, @batpad – https://github.com/IFRCGo/go-frontend/pull/2198/files

szabozoltan69: We call o.country_to.name in case the country_to does exist – the row begins with this condition: "o.country_to ? ..."

PR #2224, opened 2022-02-11
Slug usage in Emergency page

szabozoltan69:


szabozoltan69:
Works fine:
kép

Though on backend there is still no "nice" slug-creating admin interface.

PR #2241, opened 2022-02-23
fix - config.js

vdeak:


szabozoltan69:
Fixed via another branch

Issue #2212, opened 2022-02-04
Redirect 'region' country pages to region pages


tovari:
## Issue
We have regions in the country table as some emergencies are linked to regions. So regions have country pages as well. It is very confusing, because those country pages of Eurpoe, MENA, etc. seem like the region pages, with much less content. We shouldn't use these 'region' country pages on the UI, and these 5 pages should be redirected to the region pages. Users will see always the standard region page.

E.g. Europe country page 
https://go.ifrc.org/countries/286#operations
should be redirected to
https://go.ifrc.org/regions/3#operations

PR #2249, opened 2022-02-24
Redirect region semi-countries to regions

szabozoltan69:
Solves https://github.com/IFRCGo/go-frontend/issues/2212

### DIFF
--- a/src/root/views/Country/index.js
+++ b/src/root/views/Country/index.js
@@ -1,6 +1,6 @@
 import React from 'react';
 import memoize from 'memoize-one'; import { PropTypes as T } from 'prop-types';
-import { Link } from 'react-router-dom';
+import { Link, Redirect } from 'react-router-dom';
 import { connect } from 'react-redux';
 import { DateTime } from 'luxon';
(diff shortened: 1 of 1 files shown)


batpad:
@szabozoltan69 nice !

We should see if we can solve this to have the region id specified in the database or something so we don't need to hard-code these values on the frontend. That does seem like it might take some thinking. For now, can we put the country_id -> region_id mapping just in one variable, so it's in one place, so something like:

    const countryRegionMapping = {
      '282': '1',
      '283': '2',
      ... 
    }, etc.

And then like:

    if (this.props,.country.id in countryRegionMapping) {
      return (<Redirect to=`/regions/${countryRegionMapping[this.props.country.id]}` />) 
    }

^ something like that.
Would just make it easier to follow and change in the future.

szabozoltan69:
Thanks @batpad ! I fix this...

szabozoltan69:
Much nicer: https://github.com/IFRCGo/go-frontend/pull/2250/files – thanks @batpad

Issue #2261, opened 2022-03-01
[PROD] Name shows as Undefined in Deployments popup for logged out users


batpad:
## Issue

In the popup when you hover over a Deployment in the Deployments table, for eg. on https://go.ifrc.org/emergencies/5854#surge - if you are not logged in, you see the word `Undefined` instead of the name of the deployed person.

## Steps to reproduce

Go to https://go.ifrc.org/emergencies/5854#surge . Make sure you are logged out. Hover over a line in the Deployments table. See popup - there is a text saying `Undefined` instead of the deployed person's name.

## Expected behaviour

We don't want to show names to logged out users, so this is fine - we should just not show the text `Undefined`. @tovari @anamariaescobar do you know what we should show there instead? Just a `-` or a message saying `Name Hidden` or just `***` .. or ... ?

This should be a fairly easy fix, need to handle there being no name here: https://github.com/IFRCGo/go-frontend/blob/develop/src/root/components/connected/personnel-table.js#L220

cc @szabozoltan69

szabozoltan69:
Once I fixed a similar bug, but not on the "mouse hover", but somewhere else.
Here the solution was *** instead of the name. To be consequent, we could use it also.
https://github.com/IFRCGo/go-frontend/commit/08ff0efd

anamariaescobar:
I agree with @szabozoltan69 solution, *** to keep it consistent :)

batpad:
Fixed in 5.8.2 as a hotfix. @tovari this can probably be closed after the fix is verified.

PR #2267, opened 2022-03-01
make consistent the string that shows for undefined name, refs #2261

batpad:
Refs #2261 

cc @szabozoltan69


szabozoltan69 (approved):
Clean, clear!

Issue #2216, opened 2022-02-08
[PROD] FR title should remain in 100 characters


szabozoltan69:
There was a FR which had this title:
_Movilizaciones ciudadanas contra la migración desregulada en las regiones de Tarapacá y de Arica y Parinacota._ (111 characters)

This tried to be the name of the new event (emergency). This emergency did not exist before. The size of the used "name" field in database: _character varying(100)_

So this FR creation gave this error, due to not being able to create the attached event:
DataError at /api/v2/create_field_report/ – value too long for type character varying(100)

The error is the too long field. It should be handled on frontend side, not allowing (or trimming) the too long text.
Another solution possibility (if really needed to have such long title) is to make a bigger database field.

szabozoltan69:
@tovari @batpad @luisfanovich what do you think about the better solution?

tovari:
I don't know, why it is limited to 100 characters. Is there maybe a UI limitation, i.e. the space for showing the FR name is limited somewhere on the UI?
The coming FR nomenclature will bring a bit longer FR names as ISO3 code, disaster type, date will be included, so maybe it would worth to increase the number of characters. And the UI should validate the entry and warn the user if it is too long.

szabozoltan69:
I also do not know the reason – the fact is simply that: the database field is a char(100).
We can change it to text (to have no limitations).

szabozoltan69:
I limited it on client side. The too long titles are to be avoided, it's hard to read them. Let's be concise.

szabozoltan69:
Tested locally: not allows more, than 100 characters.

tovari:
The length is now limited to 250 characters, right @szabozoltan69 ?

szabozoltan69:
@tovari
It's true, that on backend side we changed it to 256:
https://github.com/IFRCGo/go-api/blob/develop/api/models.py#L552
- but on frontend side the developer forgot to release the 100 characters limit:
https://github.com/IFRCGo/go-frontend/blob/develop/src/root/views/FieldReportForm/ContextFields/index.tsx#L269

PR #2317, opened 2022-04-01
Allowing longer FR title

szabozoltan69:


szabozoltan69:
Solves https://github.com/IFRCGo/go-frontend/issues/2216

PR #2339, opened 2022-04-15
Global3w - also IFRC, ICRC in 3W NS list

szabozoltan69:


batpad:
We should have a better way of doing this than hard-coding the ISO valus here.

@szabozoltan69 - of course, this is great to immediately fix the issue, but just tagging @jhenshall and @geohacker here to keep this on the agenda to discuss this special case and how we can achieve this without hard-coding individual ISO codes.

szabozoltan69:
Good catch, @batpad . We could have one simple (new) flag for this purpose in the country table. (Otherwise the Benelux states is hard to find without any literal string.)

PR #2342, opened 2022-04-19
tinyApikey + env values

szabozoltan69:
Let's consolidate variables


szabozoltan69:
Done via another change

PR #2343, opened 2022-04-25
Add RichTextOutput component for sanitized html output

frozenhelium:


frozenhelium:
@batpad Here's the safe component to output the HTML

PR #2352, opened 2022-04-29
Export emergency response activities - v1.0

szabozoltan69:
A nicer solution would be needed, not using a copy of components/export-button-container.js as components/export-button-container2.js, but using some parameters.

Copying of components/export-button-container2.js happened due to the URL change, which is not ".../?format=csv" but "...&format=csv" (without a / , with & instead of ?).


szabozoltan69:
kép

szabozoltan69:
@frozenhelium Do you have any suggestion about how could we avoid having a separate components/export-button-container2.js which is just a slight change of components/export-button-container.js ?

szabozoltan69:
Please have a look at the recent version, dear @frozenhelium 
I introduced an onlyFromAndSign switch.
Maybe I shall write it some other part of [src/root/components/export-button-container.js](https://github.com/IFRCGo/go-frontend/pull/2352/files#diff-d83662952374295276a5da9b46ee645b6a572dc8fc63d45783bdff6569bff8d7) also.

PR #2363, opened 2022-05-13
Flash update export

frozenhelium:


szabozoltan69:
Should we deploy it via the new pipeline (to staging, https://go-stage.ifrc.org )?
(The automatic deployment does not work any more.)

Issue #2367, opened 2022-05-24
Capitalize GO in page titles


tovari:
## Issue
Please make capital GO in all page titles (prod and stage as well):

![image](https://user-images.githubusercontent.com/68283148/170062194-50239744-4f0e-45c4-a2a0-0fe39883e3f0.png)

Should be IFRC GO....

szabozoltan69:
Changed all IFRC Go strings to capital: IFRC GO.
Maybe a translation reload is needed to take effect, because the translation keys did not change (just the value).

Issue #2368, opened 2022-05-24
Response Documents search on the emergency page Reports tab


tovari:
## Issue

The search field of the Response Documents is case sensitive, which makes search difficult. It should be case insensitive.

E.g.:
https://go.ifrc.org/emergencies/5854#reports

![image](https://user-images.githubusercontent.com/68283148/170066668-1268089a-1d4a-4ebb-9d83-cb2a81d14ee5.png)

![image](https://user-images.githubusercontent.com/68283148/170066786-3013d667-9cff-4ede-8494-511c42573a6c.png)

szabozoltan69:
Solution found:
![kép](https://user-images.githubusercontent.com/6050643/170103230-ebee0832-c223-4e3a-80e0-c2b4de25b243.png)

PR #2372, opened 2022-05-24
Capitalize IFRC GO

szabozoltan69:
Changed all IFRC Go strings to capital: IFRC GO.
Maybe a translation reload is needed to take effect, because the translation keys did not change (just the value).
(I mean: /translation-dashboard, updated by dev, ...  - or reload via table export/import, even with litteral "IFRC Go->IFRC GO" replacement in other translations as well.)

Fixes https://github.com/IFRCGo/go-frontend/issues/2367


szabozoltan69:
Thanks to https://www.tutorialspoint.com/javascript-filter-an-array-of-strings-matching-case-insensitive-substring this is also solved: https://github.com/IFRCGo/go-frontend/issues/2368

PR #2375, opened 2022-05-27
Placeholder to TinyMCE

szabozoltan69:
According to https://www.tiny.cloud/blog/tinymce-placeholder-text


szabozoltan69:
@frozenhelium Why is it possible to change a "const" variable this way? I am surprised that it works...
https://github.com/IFRCGo/go-frontend/blob/feature/placeholder-to-tinymce/src/root/components/RichTextArea/index.tsx#L11
and
https://github.com/IFRCGo/go-frontend/blob/feature/placeholder-to-tinymce/src/root/components/RichTextArea/index.tsx#L61

Issue #2307, opened 2022-03-25
NS Requests International Assistance - on emergencies


szabozoltan69:
The ns_request_assistance field is in the fieldreport data table.
The emergency pages (e.g. https://go.ifrc.org/emergencies/5854 ) show an **NS Requests International Assistance** also, but based on...       recently **a** fieldreport:
https://github.com/IFRCGo/go-frontend/blob/develop/src/root/components/emergencies/overview.js#L66

We should think over it. Really a random (first) fieldreport is ok? Or should we organize it by an other way, e.g. if there is one fieldreport which has ns_request_assistance = true, then the whole emergency should be "NS Requests International Assistance" YES?

szabozoltan69:
@tovari @batpad

szabozoltan69:
Thanks to @anamariaescobar – let's select only those fieldreports, that concerns to the emergency, ant let's take the last one, and let it define the shown "NS Requests International Assistance".

PR #2378, opened 2022-05-31
NS Requests International Assistance - on emergencies

szabozoltan69:
For https://github.com/IFRCGo/go-frontend/issues/2307


szabozoltan69:
@frozenhelium This could be solved via filtering instead of pushing the needed objects into an array, isn't it?
But how?

Issue #2359, opened 2022-05-06
[PROD] Tab visibility


szabozoltan69:
## Issue
If we call https://go.ifrc.org/emergencies/3380 - that is the first view as superuser:
![kép](https://user-images.githubusercontent.com/6050643/167126672-318e9cf6-8555-42f3-824f-a4274538ba62.png)
If we click to Surge tab, after a bit of searching it disappears:
![kép](https://user-images.githubusercontent.com/6050643/167126782-c6afeb1a-c851-43e2-a62a-98364dccb0ae.png)
The Additional Information tab shows this:
"This content is not available" (PowerBI)
![kép](https://user-images.githubusercontent.com/6050643/167126873-b6cf9cb7-f0e0-4d57-909e-25949add268a.png)

We should fix this. Superuser should be able to see everything.

szabozoltan69:
The root cause is the difference between the two queries:
- one for listing tabs (does **not** use is_active=true)
- one after clicking to the tab (uses is_active=true)
So the two queries:
https://api.go.ifrc.org/api/v2/surge_alert/?event=3380&limit=5&offset=0
https://api.go.ifrc.org/api/v2/surge_alert/?limit=5&ordering=is_stood_down&is_active=true&event=3380&offset=0
The first gives count=1, the second: count=0, due to 3380 event is not is_active.

PR #2380, opened 2022-05-31
Emergency hasRRTab needs is_active surge alert

szabozoltan69:
-    this.props._getSurgeAlerts(1, { event: id });
+    this.props._getSurgeAlerts(1, { 'event': id, 'is_active': true });


szabozoltan69:
Heals https://github.com/IFRCGo/go-frontend/issues/2359

Issue #2389, opened 2022-06-03
[PROD] Elasticsearch returns results that user isn't allowed to view


esario:
## Issue

Elasticsearch returns results that user isn't allowed to view.

## Steps to reproduce

1. Log in with an NS user account
2. Search for "Bangladesh Monsoon Floods"
3. Elasticsearch returns many results, one of them is Event: Bangladesh : Monsoon Flood 10/2019 [(front-end)](https://go.ifrc.org/reports/11966) [(backend)](https://goadmin.ifrc.org/en/api/fieldreport/11966/)
4. Select the event
5. "Resource Not Found! The resource doesn't exist or you are not authorized to access this resources" error message is shown

![image](https://user-images.githubusercontent.com/15220593/171804228-121c2eab-4e08-49dd-98f2-dc2388e04943.png)

## Expected behaviour

Search should return only results that user is allowed to view.

## Impact

Potentially every user who uses the search feature.

## Severity

9 out of 10 patients might develop slight symptoms of confusion and irritation. 1 out of 10 might show more severe symptoms of anxiety.

## Is there a workaround?

None identified.

szabozoltan69:
Recently we do not store visibility data in the ElasticSearch indexing.
We should smarten our system with this feature, but this needs some effort.

A (dirty) workaround would be an extension to the warning, like "...if you are sure that the needed page exists, please write an email to im@ with the required entity and the team will email you the needed info".

szabozoltan69:
Good news: succeeded to put a visibility value to the indexing.

Example:
![kép](https://user-images.githubusercontent.com/6050643/171907694-04c9bc4e-0a66-4917-a0f3-5d5e820c4bb4.png)
https://github.com/IFRCGo/go-api/pull/1444

So the task is to be aware in https://github.com/IFRCGo/go-frontend/blob/develop/src/root/utils/utils.js#L368-L374
with the user's profile entities (on frontend).

szabozoltan69:
Another good news: https://github.com/IFRCGo/go-frontend/pull/2394/files uses a "linear" permission level comparison and shows only those that are permitted.

szabozoltan69:
Can be tested on Staging.

tovari:
This is fine partially. Filtering of non-public content seems to work. 
On the other hand NS users may get search result of content visible to other NSs only.
E.g.:

[...]

PR #2394, opened 2022-06-07
Show only permitted search result records

szabozoltan69:
Linear, comparable permission level order:
```
  2IFRC                      9
  4IFRC_NS                   8
  1MEMBERSHIP (logged in)    7
  3PUBLIC                    6
```


szabozoltan69:
Will solve https://github.com/IFRCGo/go-frontend/issues/2389

PR #2407, opened 2022-06-24
Simple line distance in RichText description

szabozoltan69:


szabozoltan69:
@frozenhelium @batpad – Do you know a better scss placement (or maybe a not-constant-dependent) solution for this?
The test page is https://go-stage.ifrc.org/reports/10036 | check the Description please, the coloured paragraph should not be broken to lines.

szabozoltan69:
With the HTML sanitation this change can be skipped.

Issue #2177, opened 2022-01-04
Rich text support for field report narrative fields


tovari:
As a field report creator | GO content consumer, I would like to have basic rich text features available when writing (reading) field reports, so that long chunks of text would be easier to organise/read. See screenshot for a recent example that doesn't encourage reading:
![MicrosoftTeams-image (24)](https://user-images.githubusercontent.com/68283148/148094866-366b99ca-c128-4634-abef-8cd73799cfa7.png)

tovari:
Rich text editing options are available on the admin page and the formatted text can be also displayed correctly on the frontend UI.
Would be great to have on the frontend FR creation form at least the following formatting options:
- newline

tovari:
@batpad, @szabozoltan69, would you mind to check the complexity of this request and estimate the required work efforts? Thanks!

szabozoltan69:
Since there are already examples for rich text fields, it should not be very hard to implement it.
The migration of the old data could be a bit shaky, but overall I estimate it a 2 working day's task.

tovari:
Thanks @szabozoltan69, line breaks are sufficient on the frontend, I think we don't need to support other rich text formatting options there.
The normal text line breaks should be accepted and handled as line breaks, without the need of adding any line break formatting code like <br>

tovari:
We should make it available in the following text fields:
- Description
- Sources (other)
- Actions taken (other)
- All actions taken - Summary fields

batpad:
@tovari do you have some good test URLs either on staging or prod where this problem can be seen? Thanks!

tovari:
@batpad a couple with FE and BE links:
1.
https://go.ifrc.org/reports/15321
https://goadmin.ifrc.org/en/api/fieldreport/15321

2.
https://dsgoproxyapp.azurewebsites.net/reports/9966
https://dsgocdnapi.azureedge.net/en/api/fieldreport/9966/change/

Looks like the Description field loses structure formatting already with the submission of the FR. while the other fields preserve that at least on the admin page.

tovari:
@batpad , thanks for thinking through the rich text editing options. As I see, this might be a more complex and time consuming topic we might have thought.
I'm wondering, if there is a chance to improve quickly the current display of texts on the UI by preserving the line brakes? That would already improve a lot. Thanks!

batpad:
Sorry did not look at this for a bit -

[...]

Issue #2133, opened 2021-11-30
Informal Update


mmusori:
New Report type moduled on the Field Report. It is a quick alert – to inform internal stakeholders and partners that we are monitoring an evolving situation. It enables recipients to quickly grasp the potential magnitude of the event, as well as possible RCRC engagement. Informal updates are issued for disaster and crisis events that are considered to have public interest, strategic interest or operational interest. By sharing with PNS and other partners demonstrates the value-add of the Emergency Operations team as a source of critical information, and ‘unlocks’ knowledge that was previously generated and held within IFRC.
The design  and documentation are available in the prototype. 
The form: https://www.figma.com/file/u0XifdJwTEx165x3e0eoP9/IFRC-GO_UI-Prototype?node-id=3557%3A96977

The report: https://www.figma.com/file/u0XifdJwTEx165x3e0eoP9/IFRC-GO_UI-Prototype?node-id=636%3A15756

tovari:
Comments from DFS:
https://docs.google.com/document/d/1LHctkW3Q6NHoyXl_oTrPIuvuTxA7PNo2ZPQjZGp0h6M/edit

suraj-ram-shrestha:
**INFORMAL UPDATE REQUEST**

**UI SIDE CHECKLIST**
- [ ] Add "Informal Update Request" item in "Create a Report" dropdown
- [ ] Create a landing page for "Informal Update Request" when clicked on a dropdown item
- [ ] On a landing page, display title, description, sliding bar, and buttons as in wireframe
- [ ] When clicked on the tab of the slider bar, the related page should be open
- [ ] When clicked to Continue if no error on the page should go to next page
- [ ] When clicked to Back should go to the previous page
- [ ] When clicked to Save and Close, all changes should be saved and redirect to 
- [ ] On a landing page, Create a component for the "Context" page 
- [ ] Add section fields as shown in wireframe for "Context" page
- [ ] On the title field, The title must be automatically pre-populated with the country name and the hazard type
- [ ] Add Map section field to upload/drag & drop Map related image
- [ ] Add Graphic section field to upload/drag & drop the related image
- [ ] Add section fields as shown in wireframe for "Actions" page
- [ ] Add section fields as shown in wireframe for "Focal points" page

tovari:
I have got the following error after submission:
[...]

PR #2420, opened 2022-07-19
Adding more HTML tags for TinyMCE

szabozoltan69:
Refers to https://github.com/IFRCGo/go-frontend/issues/2177
Also refers to https://github.com/IFRCGo/go-frontend/issues/2133

### DIFF
--- a/src/root/components/RichTextOutput/index.tsx
+++ b/src/root/components/RichTextOutput/index.tsx
@@ -10,18 +10,20 @@ function useSanitizedHtml(rawHtml: string) {
       rawHtml,
       {
         allowedTags: [
-          'b',
-          'i',
-          'em',
-          'h',
-          'p',
-          'bold',
-          'strong',
-          'li',
-          'ul',
-          'a',
-          'span',
-          'div',
+            // https://www.semrush.com/blog/html-tags-list
+            'p', 'br', 'hr', 'span', 'div',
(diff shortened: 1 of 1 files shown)


src/root/components/RichTextOutput/index.tsx
```diff
@@ -10,18 +10,20 @@ function useSanitizedHtml(rawHtml: string) {
...
+            'strong', 'b', 'i', 'em', 'u',
+            'li', 'ul', 'ol',
+            'a',
+            'table', 'thead', 'tbody', 'th', 'tr', 'td',
+            'dd', 'dt', 'dl',
+            'sub', 'sup',
+            'img', 'svg',
+            'pre', 'cite', 'code', 'q',
+            'base', 'iframe', 'canvas', 'video', // really needed?
+            'area', 'map', 'label', 'meter', // really needed??
```
frozenhelium, line 25: We should try to keep this one as minimum as possible. Also we need to be more careful about `iframe`

szabozoltan69: Besides the "as minimum as possible" we should consider the "user perspective" also, which is mentioned here: https://github.com/IFRCGo/go-frontend/issues/2133#issuecomment-1184167219

frozenhelium (approved):

PR #2436, opened 2022-08-09
Fix DREF Changes

puranban:


src/root/views/DrefApplicationForm/EventDetails/index.tsx
```diff
@@ -38,6 +39,50 @@ interface Props {
   isImminentOnset: boolean;
   fileIdToUrlMap: Record<number, string>;
   setFileIdToUrlMap?: React.Dispatch<React.SetStateAction<Record<number, string>>>;
+  isSlowOnset: boolean;
+}
+
+function useSanitizedHtml(rawHtml: string) {
```
frozenhelium, line 45: let's only use 'sanitizeHtml' directly from the lib

PR #2459, opened 2022-08-18
Add annual figures to 3w projects

szabozoltan69:
Refers to https://github.com/IFRCGo/go-api/issues/1189


src/root/components/ThreeWForm/index.tsx
```diff
@@ -599,6 +600,15 @@ function ThreeWForm(props: Props) {
                 value={value.status ? statuses[value.status] : undefined}
               />
             </div>
+            <div>
+              <Switch
+                  label="Annual Reporting"
+                  name="is_annual_report"
+                  value={true}
+                  onChange={true}
+                  invertedLogic
```
frozenhelium, line 609: Just a note: `intervtedLogic` prop will invert the behaviour, i.e. Switch will be off when the value is true and on if the value is false. You may not need it in this situation

szabozoltan69: Thank you – it was just a yesterday's quick copy of Switch to the needed place.

PR #2471, opened 2022-09-09
Calculating and conditionally-disabling splitted totals

szabozoltan69:
Annual split magic

.../three-w/2783 - a splitted project on staging
.../three-w/1446 - a non-splitted project on staging


szabozoltan69:
Refers to https://github.com/IFRCGo/go-api/issues/1189

PR #2519, opened 2022-10-08
DREF: Add docx import in client side

frozenhelium:


src/root/views/DrefApplicationForm/import.ts
```diff
@@ -0,0 +1,573 @@
...
+}
+
+function getBooleanSafe(str: string | undefined) {
+  if (str === 'Yes') {
+    return true;
+  }
+
+  if (str === 'No') {
+    return false;
+  }
```
tnagorra, line 104: Should we do a case insensitive check?

src/root/views/DrefApplicationForm/import.ts
```diff
@@ -0,0 +1,573 @@
...
+
+  return undefined;
+}
+
+function getDateSafe(str: string | undefined) {
+  if (isNotDefined(str)) {
+    return undefined;
+  }
+
+  const date = new Date(str);
```
tnagorra, line 114: While we are doing this we need to check if we want to add 'time' to the 'date' string

src/root/views/DrefApplicationForm/import.ts
```diff
@@ -0,0 +1,573 @@
...
+    return undefined;
+  }
+
+  if (trimmedStr === 'Enter indicator title.') {
+    return undefined;
+  }
+
+  if (trimmedStr === 'Name.' || trimmedStr === 'Email.' || trimmedStr === 'Title.' || trimmedStr === 'Phone number.') {
+    return undefined;
+  }
```
tnagorra, line 143: Let's use a map instead. Also, it would be best if we could add some prefix/suffix on the document itself.

src/root/views/DrefApplicationForm/import.ts
```diff
@@ -0,0 +1,573 @@
...
+  ];
+  const needs_identified = needItems.map((needItem) => {
+    const order = +(needItem.key.substring(
+      NEED_KEY.length,
+      needItem.key.length,
+    ));
+
+    if (Number.isNaN(order) || isNotDefined(order) || !needItem.value) {
+      return undefined;
+    }
```
tnagorra, line 433: Why not just use the getNumberSafe function

src/root/views/DrefApplicationForm/import.ts
```diff
@@ -0,0 +1,573 @@
...
+  ];
+  const national_society_actions = nsActionItems.map((nsActionItem) => {
+    const order = +(nsActionItem.key.substring(
+      NS_ACTION_KEY.length,
+      nsActionItem.key.length,
+    ));
+
+    if (Number.isNaN(order) || isNotDefined(order) || !nsActionItem.value) {
+      return undefined;
+    }
```
tnagorra, line 389: Why not just use the getNumberSafe function

src/root/views/DrefApplicationForm/import.ts
```diff
@@ -0,0 +1,573 @@
...
+      const indicatorItems = getItemsWithMatchingKeys(keyValueList, INDICATOR_KEY);
+      const groupedIndicators = listToGroupList(
+        indicatorItems,
+        (d) => {
+          const tempKey = d.key.substring(INDICATOR_KEY.length, d.key.length);
+          return tempKey.split('.')[0];
+        },
+        (d) => {
+          const tempKey = d.key.substring(INDICATOR_KEY.length, d.key.length);
+          const newKey = tempKey.split('.')[1];
```
tnagorra, line 298: I think we need a regex or a utility function for this operations.

src/root/views/DrefApplicationForm/import.ts
```diff
@@ -0,0 +1,573 @@
...
+  const interventionItems = getItemsWithMatchingKeys(importData, INTERVENTION_KEY);
+  const groupedInterventions = listToGroupList(
+    interventionItems,
+    (d) => {
+      const splits = d.key.split('_');
+      return splits[splits.length - 1];
+    },
+    (d) => {
+      const lastIndex = d.key.lastIndexOf('_');
+      const newKey = d.key.substring(INTERVENTION_KEY.length, lastIndex);
```
tnagorra, line 256: I think we need a regex or a utility function for this operations.

src/root/views/DrefApplicationForm/index.tsx
```diff
@@ -486,6 +496,36 @@ function DrefApplication(props: Props) {
 
   const failedToLoadDref = !pending && isDefined(drefId) && !drefResponse;
 
+  const handleDocumentImport = React.useCallback(async (newValue: File | undefined) => {
+    if (!newValue) {
+      return;
+    }
```
tnagorra, line 502: Let's show some error.

src/root/views/DrefApplicationForm/index.tsx
```diff
@@ -486,6 +496,36 @@ function DrefApplication(props: Props) {
...
+
+    const importData = await getImportData(newValue);
+    if (importData) {
+
```
tnagorra, line 506: We should have try catch and also check if the component is still mounted

tnagorra (changes requested):

src/root/views/DrefApplicationForm/import.ts
frozenhelium: It's a select field in the document

src/root/views/DrefApplicationForm/index.tsx
frozenhelium: We don't need to, this is in case the value selected was not defined

PR #2523, opened 2022-10-12
Fix operational update new changes

puranban:


src/root/views/DrefOperationalUpdateForm/Overview/index.tsx
```diff
@@ -1,10 +1,10 @@
...
-  EntriesAsList,
-  Error,
-  getErrorObject,
-  PartialForm,
-  SetBaseValueArg,
+    EntriesAsList,
+    Error,
+    getErrorObject,
+    PartialForm,
+    SetBaseValueArg,
```
frozenhelium, line 7: We should use 2 spaces

frozenhelium (approved):

PR #2529, opened 2022-10-18
First draft for using a fixed country list

szabozoltan69:


src/root/components/navbar.js
```diff
@@ -46,6 +46,31 @@ function Navbar(props) {
   const loadOptionsWithDebouncing = React.useCallback((input, callback) => {
     window.clearTimeout(debounceTimeoutRef.current);
 
+    let ans = [];
+    if (input.length < 3) {
```
szabozoltan69, line 50: @frozenhelium @batpad could you please suggest me another condition (instead of input.length < 3) that would be true when the next request is not fulfilled?

src/root/components/navbar.js
```diff
@@ -46,6 +47,18 @@ function Navbar(props) {
   const loadOptionsWithDebouncing = React.useCallback((input, callback) => {
     window.clearTimeout(debounceTimeoutRef.current);
 
+    if (input.length == 2) {
```
szabozoltan69, line 50: @frozenhelium @batpad could you please suggest me another condition (instead of input.length == 2) that would be true when the next request is not fulfilled?

src/root/components/CountryMatches.js
```diff
@@ -0,0 +1,116 @@
+export const CountryMatches = {
```
batpad, line 1: @szabozoltan69 sorry am just looking at this, but can we please please try and avoid hard-coding this list like this? We should just be able to get the countries from the redux store and then loop through them to search by substring?

szabozoltan69: @batpad We can find an other solution, yes, in the next release :-) – but then we should implement some logic also with it, which is not straightforward. E.g. the 'United Arab Emirates' can be found now with 'un', 'ar', or with 'em'. But this rule (beginning of words' is not always works, because there are a lot of "Republic..."-s or "Kingdom..."-s).
And there are other tricks like this, e.g. "us" -> United States, "uk" -> United Kingdom etc.
So, for now, it seemed to be the more simple way to create this collection to have a relevant search possibility.

My hardest question is to have a better solution instead of this: 2 letters -> wired in list, 3 letters -> ElasticSearch.
Instead of we would like to know that "the answer still not arrived, in the meantime we show this".

Issue #2492, opened 2022-09-21
Having country "result" at once in search bar


szabozoltan69:
There is a complaint for having too much time until the search results appear, even when searching for countries.

The country pages mean 192 pages. We could have these 192 links in frontend literal memory and if the searched string matches some/one of these, we display them at once, and then, when in 1-2 seconds the other results arrive, we put them after these.

So the user can have the experience of "typing 3 letters" of a country and at once there is a result of the country page, and if the user has a bit more patience, the others also appear below.

@mariam-yu @frozenhelium

szabozoltan69:
Let's use useReduxState('countries')
And maybe we could showing that collecting results are in progress.

szabozoltan69:
@frozenhelium @tnagorra @batpad 
I would like to ask some help in this topic.
Where to put the "fixed country list"?
I tried this way: replacing this row: https://github.com/IFRCGo/go-frontend/blob/develop/src/root/utils/utils.js#L429
with some fixed thing, now (for simplicity) let's use this only:
```
      const fixes = [{ value: "/countries/14", label: "Country: Afghanistan" }];
      callback(fixes.concat(options));
```
But this does not help, because the callback gives back the answer when the whole options list is arrived.
How could we have a "quick" result? Maybe somewhere https://github.com/IFRCGo/go-frontend/blob/develop/src/root/components/navbar.js#L50 here where getElasticSearchOptions is called?

szabozoltan69:
@frozenhelium @tnagorra @batpad 
I would like to ask some help in this topic.
Where to put the "fixed country list"?
I tried this way: replacing this row: https://github.com/IFRCGo/go-frontend/blob/develop/src/root/utils/utils.js#L429
with some fixed thing, now (for simplicity) let's use this only:
```
      const fixes = [{ value: "/countries/14", label: "Country: Afghanistan" }];
      callback(fixes.concat(options));
```
But this does not help, because the callback gives back the answer when the whole options list is arrived.
How could we have a "quick" result? Maybe somewhere https://github.com/IFRCGo/go-frontend/blob/develop/src/root/components/navbar.js#L50 here where getElasticSearchOptions is called? Or can we use an other condition instead of request...then(), some kind of ...else (in the case where the "then" is not yet actual)?

szabozoltan69:
[...]

PR #2534, opened 2022-10-25
No wired-in country list for search acceleration

szabozoltan69:
@batpad Many thanks for this intelligent and solid, concise solution!


szabozoltan69:
Refers to https://github.com/IFRCGo/go-frontend/issues/2492

src/root/components/navbar.js
```diff
@@ -48,23 +48,41 @@ function Navbar(props) {
...
+        return country.name.toLowerCase().startsWith(input.toLowerCase())
+            && country.independent
+            && !country.is_deprecated;
+      }).concat(exceptionCountries).map(country => {
+        return {
+          value: `/country/${country.id}`,
+          label: `Country: ${country.name}`
+        };
+      });
+      //if (input in CountryMatches) {
```
batpad, line 70: @szabozoltan69 can we clear out commented code?

szabozoltan69: Yes, I cleared it (and used "countries" instead of "country").
https://github.com/IFRCGo/go-frontend/blob/develop/src/root/components/navbar.js#L70

PR #2542, opened 2022-11-08
Tooltips for CSV exports

szabozoltan69:
@frozenhelium @tnagorra Please advice me a better solution for title assignment than this renderToStaticMarkup + replace. (Without importing the full lang "strings", of course.)


szabozoltan69:
Refers to https://github.com/IFRCGo/go-api/issues/1430

Issue #2557, opened 2022-12-02
[PROD] Surge - Catalogue of services - text update


tovari:
Please change the text in the Catalogue that in the Health section on the Rapid Response personnel, we have “Emergency Relief Unit Role profiles....” It should be Emergency Response Unit Role Profile instead...

![image](https://user-images.githubusercontent.com/68283148/205266661-aada606d-f547-4d58-bdf3-c228702e9559.png)

szabozoltan69:
So in fact this change is required (with plural, the 2nd one), I suppose:

Emergency Relief Unit Role profiles
->
Emergency Response Unit Role Profiles

tovari:
https://go-stage.ifrc.org/deployments/catalogue/health

szabozoltan69:
Deployed to Staging.

PR #2558, opened 2022-12-02
ERU new name is Emergency Response Unit - text fix

szabozoltan69:


szabozoltan69:
Refers tp https://github.com/IFRCGo/go-frontend/issues/2557

Issue #2555, opened 2022-11-30
[PROD] "Explore 3W Regional" button leads to a blank page


steph-gm:
## Issue
On the 3W section, when scrolling to the map, we see the "Explore 3W Regional" button, and when clicking on it we see a list with the different regions. There, when selecting and clicking any of the regions, the page starts loading, but after a few seconds it will just show a blank page.
This issue was found late november 2022. I got the same blank page from 2 different browsers (chrome and firefox) and from different computers as well.
- Link to 3W main page: https://go.ifrc.org/three-w/
- One of the links not working (copy pasted from the url bar after selecting the Americas region): https://go.ifrc.org/regions/1#3w

## Steps to reproduce
1. Go to the 3W page/section (URL: https://go.ifrc.org/three-w/)
2. Scroll to the "NS WITH ONGOING PROJECTS" map and click on the "Explore 3W Regional" button
3. In the displayed list, select and click on any region name
The page will not load. And when going back, netiher the main 3W page.

## Expected behaviour
Page should show the 3W for the specific region that was selected

## Impact
Affects anyone looking at the 3W page and wanting to explore the regional 3W page. 

## Severity
Minimal impact. Prevents of using the option, but this has not been reported by any user, it was noted when reviewing the 3W functionality

Note: This is also hapenning with Staging.

szabozoltan69:
The error message in development environment:
![image](https://user-images.githubusercontent.com/6050643/205023018-11d68b25-8bb6-4007-bdbd-adbb74ce0b25.png)

szabozoltan69:
It seems to be the problem: the ownPorps.match.params.id contains not only the id but the #string also, so e.g. 0#3w.
It is possible to cut it down (via a split('#')[0] ), but this way I get an URL like  0#3w#operations, so an automatic appending of the default tab (#operations) appears.

szabozoltan69:
The normal Region pages are also blank at first usage (only get shown after reloading).

PR #2564, opened 2022-12-05
Quick fix for 3w regions blank page

szabozoltan69:
@batpad @frozenhelium @tnagorra – if you have better solution, please tell me.
Refers to https://github.com/IFRCGo/go-frontend/issues/2555

### DIFF
--- a/src/root/views/region.js
+++ b/src/root/views/region.js
@@ -252,6 +252,10 @@ class AdminArea extends SFPComponent {
       this.props.history.replace(`${url}${tabHashArray[index]}`);
     };
     const hashes = tabDetails.map(t => t.hash);
+    if (this.props.location.hash === '#3w#operations') { // FIXME! Should not happen! A quick & dirty hack for go-frontend/issues/2555
+        this.props.location.hash = '#3w';
+        this.props.history.replace(`${this.props.location.pathname}${this.props.location.hash}`);
+    }
     const selectedIndex = hashes.indexOf(this.props.location.hash) !== -1 ? hashes.indexOf(this.props.location.hash) : 0;
 
     const foldLink = (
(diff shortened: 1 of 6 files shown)


frozenhelium:
@szabozoltan69 The right approach would be to add a `/` at the end of the regions link in the 3W page. This approach reveals another mistake in the Multiplexer. We've set the route of Regions to match with `exact` which is not desired behaviour (Since, we have the hash). I've fixed it now, please do review :)

frozenhelium (approved):

frozenhelium (approved):

src/root/views/region.js
```diff
@@ -568,7 +568,7 @@ const selector = (state, ownProps) => ({
   appealsListStats: state.overallStats.appealsListStats,
   countriesByRegion: countriesByRegionSelector(state),
   regions: regionsByIdSelector(state),
-  thisRegion: regionByIdOrNameSelector(state, ownProps.match.params.id),
+  thisRegion: regionByIdOrNameSelector(state, ownProps.match.params.id.split('#')[0]),
```
szabozoltan69, line 571: @frozenhelium should I remove this cutdown also?

frozenhelium:
@szabozoltan69 turns out it was the problem with the link fix, I've pushed with the appropriate fix now

src/root/views/region.js
frozenhelium: I did it in my latest push :D

frozenhelium (approved):

szabozoltan69:
Beautiful and clean solution, thank you so much, @frozenhelium!

szabozoltan69:
Deployed to Staging.

Issue #2563, opened 2022-12-05
Wiki - add Risk module hook points


tovari:
We need a few new hook points on the UI related to the Risk watch pages.
The pointers show in the screenshots where to place the ? icons.

1.  Regional pages Risk tab. E.g.:
https://go-stage.ifrc.org/regions/3#risk-watch
![image](https://user-images.githubusercontent.com/68283148/205617099-0c52bac7-8ca8-4daf-82c8-d821407c82c7.png)
url: 
https://go-wiki.ifrc.org/en/user_guide/risk_module

2. Regional pages Risk tab, Imminent tab. E.g.:
https://go-stage.ifrc.org/regions/3#risk-watch
Only when the Imminent toggle is active:
![image](https://user-images.githubusercontent.com/68283148/205617673-cae20617-c16f-4a0d-a1ec-a9ee29855448.png)
url:
https://go-wiki.ifrc.org/en/user_guide/risk_module#imminent-events

3. Regional pages Risk tab, Seasonal tab. E.g.:
https://go-stage.ifrc.org/regions/3#risk-watch
Only when the Seasonal toggle is active:
![image](https://user-images.githubusercontent.com/68283148/205618197-048cc643-157c-49ab-9758-7dd0d0c54472.png)
url:
https://go-wiki.ifrc.org/en/user_guide/risk_module#seasonal-risk

4. Country pages Risk tab. E.g.:
https://go-stage.ifrc.org/countries/101#risk
![image](https://user-images.githubusercontent.com/68283148/205618709-302c5a01-68e1-497e-a771-aa4be8791936.png)
url:
https://go-wiki.ifrc.org/en/user_guide/risk_module#seasonal-risk

PR #2565, opened 2022-12-05
Wiki - Risk Module links

szabozoltan69:
Refers to https://github.com/IFRCGo/go-frontend/issues/2563
Should be done much nicer, with translations usage and better (sharper) right alignment.
This way the right alignment depends on header text length, which is not acceptable.


frozenhelium:
@szabozoltan69 I've done some refactoring here, please do take a look. Thanks for the typo fix for 'Payload'.

frozenhelium:
Also, this is a very good approach (using a dedicated component for wiki link), we should generalize the component and follow similarly for other places where we have used the wiki link directly.

frozenhelium (approved):

szabozoltan69:
Thanks also for the "sub" fixes!
Deployed to Staging.

PR #2658, opened 2023-02-23
Use gitRevision plugin to show version and git hash

szabozoltan69:
I could not solve it via ReactTooltip. If you have a better way, let me know. This way I used simply the "title".
@frozenhelium ?


tnagorra (approved):

frozenhelium (approved):

frozenhelium:
@szabozoltan69 It's alright to use `title` for now!

PR #2666, opened 2023-03-09
Fix latest feedback in search and highlighted operations

barshathakuri:


src/root/components/Link/index.tsx
```diff
@@ -27,6 +27,7 @@ function Link(props: Props) {
     ...otherProps
   } = props;
 
+  console.warn('classname', classNameFromProps);
```
frozenhelium, line 30: We should remove this

src/root/views/Search/ProjectTable/styles.module.scss
```diff
@@ -23,3 +23,9 @@
     }
   }
 }
+
+.project-date {
+  display: flex;
+  gap: $spacing - 10;
```
frozenhelium, line 29: This probably wouldn't work

frozenhelium (changes requested):

frozenhelium (approved):

PR #2736, opened 2023-04-21
Non-mandatory districts in (emergency)projects

szabozoltan69:
Refers to https://github.com/IFRCGo/go-api/issues/1755


src/root/components/ThreeWForm/useThreeWOptions.ts
```diff
@@ -137,7 +137,7 @@ export const schema: FormSchema = {
       primary_sector: [requiredCondition],
       programme_type: [requiredCondition],
       project_country: [requiredCondition],
-      project_districts: [requiredListCondition],
+      project_districts: [],
```
frozenhelium, line 140: Let's also verify that backe-end accepts empty value or empty list for this, If we want to send the empty list, we have to add defaultEmptyListCondition or something like that exists

frozenhelium (approved):

PR #2741, opened 2023-04-22
Using only needed districts in country

szabozoltan69:
Refers to https://github.com/IFRCGo/go-api/issues/1755


szabozoltan69:
A nicer solution would be appreciated to highlight all districts in a country (without the requirement of giving them one by one).

PR #2742, opened 2023-04-26
UI component - Card

shreeyash07:


src/root/views/GoUI/examples/Cards/styles.module.scss
```diff
@@ -0,0 +1,12 @@
+.cards-collection {
+    display: flex;
+    flex-direction: column;
+    padding: 2rem;
```
priyesh777, line 4: Please use the newly defined css variables

priyesh777, line 4: Please use the newly defined css variables

samshara (approved):

PR #2757, opened 2023-05-02
Expandable table component

puranban:
### changes
  * Add an expandable component to the table.


samshara (approved):

src/root/components/Table/useRowExpansion.tsx
```diff
@@ -63,27 +54,25 @@ function useRowExpansion<D, K>(
             key={`${rowKey}-expanded`}
             className={expansionRowClassName}
           >
-            <td
-              colSpan={columns.length}
+            <tr
```
tnagorra, line 57: tr should not be nested inside another tr

tnagorra (changes requested):

PR #2758, opened 2023-05-02
Overlay component

samshara:


priyesh777 (approved):

src/root/views/GoUI/index.tsx
```diff
@@ -16,8 +17,13 @@ function GoUI() {
       <div>
         <Alerts />
       </div>
+      <div>
+        <Alerts />
```
shreeyash07, line 21: i think it is repeated

shreeyash07 (approved):

PR #2767, opened 2023-05-09
Refactor Button Component

samshara:
- Add changes to the button types and styles


src/root/views/GoUI/components/Button/index.tsx
```diff
@@ -1,190 +1,123 @@
...
+          {children}
+        </div>
+      )}
+      {actions && (
+        <div className={_cs(actionsClassName, styles.actions)}>
+          {actions}
+        </div>
+      )}
+    </>
+  );
```
shreeyash07, line 68: can we use ElementFragment or useBasicLayout ?

shreeyash07 (changes requested):

src/root/views/GoUI/components/Button/index.tsx
samshara: I have used `useBasicLayout` hooks for now.

shreeyash07 (approved):

priyesh777 (approved):

PR #2769, opened 2023-05-10
Add textInput component for form fields

priyesh777:
## Changes:
- Add updates to the global styles and also add updates to the imported file paths
- Also add missing core components
- Add TextInput component which will be included in the form fields

## Addresses:
- Task (TextInput): https://github.com/toggle-corp/ifrc-go-meta/issues/124#issue-1701325428

### DIFF
--- a/src/root/views/GoUI/examples/TextInputExample/index.tsx
+++ b/src/root/views/GoUI/examples/TextInputExample/index.tsx
@@ -1,8 +1,8 @@
 import React, { useCallback, useState } from "react";
-import TextArea from "../../components/TextArea";
+import TextInput from '#goui/components/TextInput';
 import styles from './styles.module.scss';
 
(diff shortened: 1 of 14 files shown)


samshara (approved):

src/root/views/GoUI/examples/TextInputExample/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+function TextInputExample() {
+    const [textGroup, setTextGroup] = useState('');
+
+    const handleTextChange = useCallback((val) => {
+        setTextGroup(val);
+    }, [setTextGroup]);
+
+    return (
+        <>
```
shreeyash07, line 13: Add header and use other example format

shreeyash07 (changes requested):

PR #2771, opened 2023-05-10
UI Components - DateInput, Checkbox, Switch

shreeyash07:


src/root/views/GoUI/components/Checkbox/styles.module.scss
```diff
@@ -0,0 +1,34 @@
+@import '~styles/settings/variables';
+
+.checkbox {
+  display: flex;
+  align-items: flex-start;
+  cursor: pointer;
+  gap: calc(var(--go-ui-spacing-md)*0.5);
```
samshara, line 7: Let's used defined CSS variables.

src/root/views/GoUI/components/Checkbox/styles.module.scss
```diff
@@ -0,0 +1,34 @@
...
+
+  .checkmark {
+    display: flex;
+    flex-shrink: 0;
+    align-items: center;
+    font-size: 1.5em;
```
samshara, line 13: Let's use defined CSS variables.

Issue #2749, opened 2023-04-28
DREF Account page enhacement


anamariaescobar:
Hello everyone!

Please find the wirreframes for the changes on the DREF account page [here:](https://www.figma.com/proto/u0XifdJwTEx165x3e0eoP9/IFRC-GO_UI-Prototype?type=design&node-id=7006-215581&scaling=min-zoom&page-id=0%3A4957&starting-point-node-id=0%3A4963)

Basically what we want to do is re-organize all the documents related to one DREF (Application, Ops Update, Final report) and group them so we can see them in a timeline like visualization. We are also replacing the word "Publish" by "Approved", the functionality will be the same. We will introduce a new page for the "Completed DREF Operations", an operation can be considered Completed once the final Report is approved.
We also want to bring the sharing functionality on the account page. The allocation form is an export that hasn't been implemented yet, if we can hide this button for now that would be great!

@udaynwa @frozenhelium @tovari @mariam-yu

udaynwa:
@anamariaescobar , thank you for creating the ticket and adding the wireframes. I have added a couple of questions/comments in the wireframe and CC-ed you. 

cc @frozenhelium @tovari

Shhhhhubh:
I tested the DREF Account page and found the following issues @puranban @udaynwa 

- [ ] Status is shown as 1 or 0 (in numbers) instead of submitted, in progress or approved
- [ ] In the filters section - "select type" and "Date range" are missing
- [ ] "View previous DREF operation" is "View Completed DREF operation"
- [ ] The created on date format is "DD-MM-YYYY" instead of "YYYY-MM-DD"
- [ ] In Final report the link bullets (The ones that should be seen after the drop down button is clicked i.e, the red bullets) are missing
- [ ] "Type of DREF" is shown as "Type of Display"
- [ ] When the last page is selected and the filters are changed the pagination sticks to the last page and shows no data
- [ ] Instead of kebab menu buttons we have meatballs menu buttons
- [ ] when the kebab menu button is clicked the view option is missing
- [ ] When the "approved" button is clicked the button should change to "Allocation form" link and this is not happening ( the button disappears this should happen only for final reports)
- [ ] In Previous DREF operations "Type of display" is being shown instead of "Submitted/Approved on"  (a completely different column )
- [ ] In Previous DREF operations the kebab menu buttons is missing
- [ ] In Previous DREF operations the "Approved" button is shown which should be "Allocation Form" also clicking the "Approved" button does nothing 
[...]

PR #2822, opened 2023-06-14
Fix Account Page

puranban:
### changes (Task: #2749)
  * Move the dropdown to create a new stage in the first row.
  * Fix the status column

### DIFF
--- a/src/root/components/DrefApplicationList/ActiveDrefList/ActiveDrefTable/index.tsx
+++ b/src/root/components/DrefApplicationList/ActiveDrefList/ActiveDrefTable/index.tsx
@@ -286,6 +286,48 @@ function ActiveDrefTable(props:Props) {
     },[getDrefId]
   );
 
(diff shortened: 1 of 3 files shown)


src/root/components/DrefApplicationList/ActiveDrefList/ActiveDrefTable/index.tsx
```diff
@@ -286,6 +286,48 @@ function ActiveDrefTable(props:Props) {
     },[getDrefId]
   );
 
+  const getInitialDropdownItems = React.useCallback(
+    (item) => {
+      const hasUnpublishedOperationalUpdate = item.unpublished_op_update_count > 0;
+      const hasUnpublishedFinalReport = item.unpublished_final_report_count > 0;
+      const isDrefLoan = item.type_of_dref === TYPE_LOAN;
+
```
frozenhelium, line 294: Let's use early return

frozenhelium (approved):

PR #2823, opened 2023-06-14
Fix Risk Watch Imminent

puranban:
### changes
  * Add severity data GDACS
  * Fix the Map footer tooltip

### DIFF
--- a/src/root/components/RiskImminentEventMap/GDACSEventMap/PointDetails/index.tsx
+++ b/src/root/components/RiskImminentEventMap/GDACSEventMap/PointDetails/index.tsx
@@ -7,7 +7,7 @@ import {
 import TextOutput from '#components/TextOutput';
 import MapTooltipContent from '#components/MapTooltipContent';
 import Link from '#components/Link';
-import { GDACSEvent, GDACSEventExposure } from '#types';
+import { GDACSEvent, GDACSEventExposure, ImminentHazardTypes } from '#types';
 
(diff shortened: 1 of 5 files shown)


src/root/components/RiskImminentEventMap/GDACSEventMap/PointDetails/index.tsx
```diff
@@ -55,11 +55,23 @@ function PointDetails(props: PointDetailsProps) {
       end_date,
       event_details,
       alert_level,
+      hazard_type
     },
     populationExposure,
     onCloseButtonClick,
   } = props;
 
+  const severityDetail = React.useCallback(
```
frozenhelium, line 64: 1. Let's use map to filter rather than using list with `includes`
2. Let's use early return

src/root/components/RiskImminentEventMap/MapFooter/index.tsx
```diff
@@ -21,6 +21,22 @@ import Container from "#components/Container";
 
 import styles from "./styles.module.scss";
 
+const gdacsPopupTitle = " Source: GDACS";
+const gdacsPopupDescription = (
+  <>
+    <p>
```
frozenhelium, line 27: Let's remove unused tags and fragment as well

frozenhelium (changes requested):

frozenhelium (approved):

Issue #2864, opened 2023-07-17
[PROD] Rapid Response Personnel timeline pop-up window


tovari:
We use to have a pop-up window when hovering the mouse on the red bar of the timeline:

![image](https://github.com/IFRCGo/go-frontend/assets/68283148/248a627e-dfc3-4830-8433-2a109aa79834)

The info pop-up is not available anymore.

udaynwa:
### Link to replicate the issue 
- https://go.ifrc.org/emergencies/6479#surge

### Info in the pop-up
- Person 
- Country 
- Date

batpad:
The last change made to this part of the code was https://github.com/IFRCGo/go-frontend/commit/de4bcc89250144fc7fd7c9180ffb9b0bc1eb6a06#diff-887937255ada86e51215a92196a4cce98c39266d7cc830514a338ed7752841deR251

Though it's a bit unclear to me how that could have broken anything.

PR #2879, opened 2023-08-08
Fix risk watch zoom and tooltip issues

puranban:
### Addresses 
- [x] #2864 
- [x] https://github.com/toggle-corp/ifrc-go-meta/issues/230
- [x] https://github.com/toggle-corp/ifrc-go-meta/issues/235

### Changes

- Fix the GDACS zoom issue
- Fix the emergency surge personnel table tooltip
- Fix the home figures card tooltip

### DIFF
--- a/src/root/components/RiskImminentEventMap/GDACSEventMap/index.tsx
+++ b/src/root/components/RiskImminentEventMap/GDACSEventMap/index.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import Map,
 {
+    MapBounds,
   MapContainer,
   MapLayer,
   MapSource,
@@ -15,10 +16,11 @@ import {
   _cs
 } from '@togglecorp/fujs';
 import {
+  point as turfPoint,
   bbox as turfBbox,
+  buffer as turfBuffer,
 } from '@turf/turf';
 import { LngLat } from 'mapbox-gl';
-import { BoundingBox, viewport } from '@mapbox/geo-viewport';
 
 import {
   defaultMapOptions,
@@ -30,6 +32,8 @@ import {
   ADAM_COLOR_GREEN,
   ADAM_COLOR_RED,
   ADAM_COLOR_CONES,
+  BBOXType,
+  fixBounds,
 } from '#utils/map';
 import {
   geoJsonSourceOptions,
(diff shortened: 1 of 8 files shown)


src/root/components/RiskImminentEventMap/GDACSEventMap/index.tsx
```diff
@@ -233,33 +239,32 @@ function GDACSEventMap(props: Props) {
...
+      const pointBuffer = turfBuffer(point, 50, { units: 'kilometers' });
 
-        return viewPort;
-      }
+      const geojson = {
+        type: 'FeatureCollection',
+        features: [
+          pointBuffer,
+          ...(footprintPositionGeoJson?.features ?? []),
+        ].filter(isDefined),
```
tnagorra, line 258: Any reason we need to filter?

src/root/views/Country/RiskWatch/ImminentEventsGDACS/index.tsx
```diff
@@ -102,6 +104,13 @@ function ImminentEventsGDACS(props: Props) {
     url: `risk://api/v1/gdacs/${eventId}/exposure`,
   });
 
+  const countryBounds = React.useMemo(
+    () => {
+      let bbox = turfBbox(country?.bbox ?? []);
```
tnagorra, line 109: Can we even pass empty array on turfBbox function?
Why are we passing bbox on turfBbox?

src/root/views/RegionalRiskWatch/ImminentEvents/ImminentEventsGDACS/index.tsx
```diff
@@ -96,6 +98,13 @@ function ImminentEventsGDACS(props: Props) {
     url: `risk://api/v1/gdacs/${eventId}/exposure`,
   });
 
+  const regionBounds = React.useMemo(
+    () => {
+      let bbox = turfBbox(region?.bbox ?? []);
```
tnagorra, line 103: Can we even pass empty array on turfBbox function?
Why are we passing bbox on turfBbox?

tnagorra (changes requested):

src/root/views/Country/RiskWatch/ImminentEventsGDACS/index.tsx
frozenhelium: We don't get the bounds format that is supported by mapbox, we transform it to correct format using turf

frozenhelium (approved):

